PR #138, opened 2026-02-19
Fix offline use case

MCarlomagno:
Restrict Rust multisig offline flow to `SwitchPsm` only and remove misleading offline paths for other proposal types:

- adds a network-only sync for offline operations (no PSM sync)
- simplifies the offline logic, restriced only to PSM switch operations
- centralizes transaction capability checks (`supports_offline_execution` / `requires_psm_ack`),


MCarlomagno:
merged into main

PR #196, opened 2026-04-23
remove reverse on signature felts

ngjupeng:
I keep getting "invalid public key commitment" every time I call multisig.executeProposal after switching from Miden testnet 0.13 to 0.14. Create / sign proposals work fine but only execute fails. After digging in, it's seems like because Miden 0.14 flipped the operand stack to little-endian, and encodeEcdsaSignatureFelts has a .reverse() which causing the error format of [...pkFelts, ...sigFelts]


ngjupeng:
I confirm that I have read and hereby agree to the OpenZeppelin Contributor License Agreement

MCarlomagno (approved):
LGTM, thanks!

PR #223, opened 2026-05-06
#218 Account lookup cold key

### LINKED ISSUE #209: Add Guardian account discovery by authorized key commitment for state recovery
## Description

Wallet recovery can start from only a seed-derived recovery/cold key. In that state, the wallet may not know the existing `account_id`, especially after Guardian switching or in the three-key model where the account ID depends on storage/auth configuration that the recovering wallet cannot reconstruct locally.

Today GET `/state requires account_id`, so a recovered wallet cannot fetch the existing Guardian state unless it already knows the account ID. Calling create again with the recovered key and a different Guardian configuration can derive a different account ID, leaving the wallet unable to query the correct Guardian record.

Add a Guardian-side discovery path that maps authorized key commitments to account IDs, likely using the account auth metadata already stored by Guardian.

### Proposed behavior:

Guardian indexes Miden accounts by authorized signer commitments, including recovery/cold key commitments where available.
A client can discover matching accounts using a key commitment, for example:
`GET /accounts?key_commitment=0x...`
or `GET /state?key_commitment=0x...`

Because one commitment may be authorized on multiple accounts, the API should return a list.
The client/wallet can then show matching accounts and let the user select which one to recover/load.
Once selected, the wallet fetches full state using the resolved `account_id`.

### LINKED ISSUE #218: feat: Support account lookup by key commitment for cold-key recovery
## Background

Miden account IDs are derived from a `storage_commitment` that includes commitments to all keys associated with the account at creation time. For 3-key accounts (cold key, hot key, Guardian key), this means the account ID cannot be derived from the cold key alone.
Problem originally raised by @0xnullifier and discussed with @bobbinth @MCarlomagno @Dominik1999.

## Problem

When a user loses their device (and therefore their hot key), the intended recovery flow is:

1. User enters seed phrase → derives cold key
2. Query Guardian for the latest account state
3. Register a new hot key

Step 2 requires the `AccountId`, but the `AccountId` cannot be recomputed from the cold key alone since it was derived from all 3 key commitments at account creation. This makes cold-key-only recovery impossible for 3-key accounts.

This is a generalization of an existing issue in the single-key case: if a user ever switched Guardian operators, the account ID derived with the new operator's key would differ from the original, requiring a workaround of always using the OZ Guardian key for ID derivation regardless of current operator.

## Proposed Solution

Guardian should maintain a mapping from key commitments to account IDs and expose a lookup endpoint, e.g.:

```
GET /state?key_commitment=<cold_key_commitment>
```

This allows a recovering user to present only their cold key commitment and retrieve their account ID (and state) without needing to know the full original key set.

### Open Questions

- **Ambiguity**: Can the same key commitment appear in more than one account? If so, this endpoint could return multiple results and the API design needs to account for that (e.g. return a list, or scope by account type).
- **Scope of mapping**: Should the Guardian index all key commitments (cold, hot, Guardian) or only cold key commitments? Indexing all would make recovery more flexible but increases storage and attack surface.
- **Mapping lifecycle on guardian switch**: When a user switches to a different Guardian operator, what should happen to the cold key/account ID mapping held by the previous operator? The previous operator's mapping becomes stale (and potentially misleading?), so it may make sense to remove it? Are there could be reasons to retain it? TBD.

## Impact

This problem is blocking the 3-key account feature release on the wallet side, as cold-key recovery is not viable without this. The fix is expected to be relatively straightforward on Guardian's side.

Image

zeljkoX:
Adds account lookup by key commitment (#218): resolves a Miden public-key commitment
   to the account IDs whose authorization set contains it, so a wallet holding only a
  signing key can recover the account ID and fetch state.
                                                                                      
  - **Server**: `GET /state/lookup` + gRPC `GetAccountByKeyCommitment`,               
  domain-separated `LookupAuthMessage` (`guardian.lookup.v1`), Postgres GIN index over
   `cosigner_commitments` with filesystem-scan fallback.                              
  - **Auth**: identity is derived from the signature itself — Falcon signatures embed 
  the public key; ECDSA signatures recover it via the recovery byte. The server then  
  enforces `commitment_of(derived_pk) == queried_key_commitment` after cryptographic  
  verification. `x-pubkey` is sent for wire-format parity with per-account requests
  but is not consulted on this path. This makes `recoverByKey` work with browser Miden
   wallets that only expose a 32-byte commitment as `publicKey`.
  - **Clients**: Rust `lookup_account_by_key_commitment` +                            
  `MultisigClient::recover_by_key`; TS `lookupAccountByKeyCommitment`,
  `Signer.signLookupMessage`, `MultisigClient.recoverByKey`. All bundled signers
  (`FalconSigner`, `EcdsaSigner`, `MidenWalletSigner`, `ParaSigner`) implement
  `signLookupMessage`.                                                                
  - **Examples**: `recover_by_key` Rust binary, demo menu action, smoke-web "Recover
  by key" button.                                                                     
  - **Parity**: shared Rust/TS digest fixtures under                  
  `crates/shared/tests/fixtures/lookup_auth_vectors.json`.
  

  ## Test plan

  - [x] `cargo test -p guardian-server -p guardian-client -p miden-multisig-client -p guardian-shared`
  - [ ] HTTP + gRPC integration tests in `crates/server/src/testing/integration/lookup_*.rs`
  - [x] `npm test` in `packages/guardian-client` and `packages/miden-multisig-client`
  - [x] Rust SDK smoke: `recover-by-key-canary` (demo `Recover by key`)
  - [x] TS SDK smoke: `recover-by-key-canary` (smoke-web button)


zeljkoX:
@0xnullifier 

Could you double-check whether these changes work for your use case?

Thanks!

crates/server/src/services/lookup_account.rs
```diff
@@ -0,0 +1,205 @@
...
+        "Lookup completed"
+    );
+
+    Ok(LookupAccountResult { accounts })
+}
+
+/// Normalize a `0x`-prefixed (or unprefixed) hex commitment to the canonical
+/// `0x`-prefixed lowercase form used in `account_metadata.auth.cosigner_commitments`.
+/// Rejects values that aren't exactly 32 bytes or contain non-hex characters.
+fn normalize_commitment_hex(hex: &str) -> Result<String> {
```
MCarlomagno, line 135: maybe there is already some shared API for this, if not I would move it to some place where it could be used across the app

MCarlomagno (approved):
LGTM, thanks!

0xnullifier:
Yup looks good, thanks!

PR #259, opened 2026-05-28
 fix(contracts): skip cleanup_pubkey_mapping when signer count does n…

### LINKED ISSUE #260: add-signer Proposal failing at proof verification
Executing add-signer proposal fails with:

```
Execute failed: failed to submit proven transaction: RPC error: grpc request failed for submit_proven_transaction: invalid request parameters: code: 'Client specified an invalid argument', message: "Invalid proof for transaction 0x4748726daa2522....: failed to verify transaction\ncaused by: failed to verify STARK proof for program with hash 0x8e3ef58005cfda....\ncaused by: constraint mismatch: quotient * vanishing != folded constraints"
````

but works for other proposals like remove signer.

zeljkoX:
Calling `exec.cleanup_pubkey_mapping` after `update_signers_and_threshold`
  unconditionally produces a STARK-invalid proof when the storage map has
  exactly one entry (e.g. add-signer from a 1-signer account). The proc's
  internal loop guard already short-circuits in this case, so the body never
  ran — but entering the proc at all perturbs the prover trace enough to fail
  node-side STARK verification with `constraint mismatch: quotient * vanishing
  != folded constraints`.

  Gate the call on `init_num_of_approvers > new_num_of_approvers` so cleanup
  only runs when the signer set actually shrinks. Add-signer and threshold-only
  updates skip the call entirely; remove-signer behavior is unchanged.
  Applied to both Falcon and ECDSA multisig MASM. The existing add-signer test
  is extended to call `prove_next_block()` so MockChain exercises proof
  generation and would catch this class of regression.

TODO: Open PR against https://github.com/OpenZeppelin/miden-confidential-contracts and verify with @onurinanc 
  

## Summary by CodeRabbit

* **Refactor**
  * Optimized multisig contract efficiency by making public key mapping cleanup conditional, now executing only when approver counts differ, with improvements applied to both standard and ECDSA contract variants

* **Tests**
  * Strengthened signer management tests by improving transaction handling and chain state progression to ensure proper validation after threshold updates

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenZeppelin/guardian/pull/259?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/contracts/masm/auth/multisig.masm
+++ b/crates/contracts/masm/auth/multisig.masm
@@ -349,14 +349,20 @@ pub proc update_signers_and_threshold(multisig_config_hash: word)
     drop
     # => [pad(12)]
 
-    # compare initial vs current multisig config
-
-    # load init_num_of_approvers & new_num_of_approvers
     loc_load.0 loc_load.1
     # => [init_num_of_approvers, new_num_of_approvers, pad(12)]
 
-    exec.cleanup_pubkey_mapping
-    # => [pad(12)]
+    dup.1 dup.1
+    u32assert2 u32lt
+    # => [should_cleanup, init_num_of_approvers, new_num_of_approvers, pad(12)]
+
+    if.true
+        exec.cleanup_pubkey_mapping
+        # => [pad(12)]
+    else
+        drop drop
+        # => [pad(12)]
+    end
 end
 
 #! Update or clear the threshold override for a specific procedure.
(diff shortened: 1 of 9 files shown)


zeljkoX:
Merged in 0.15

Raised issue against Miden protocol to fix this in upstream contracts to have it ready for #294

PR #270, opened 2026-06-03
feat: Add ECDSA AWS KMS hosted signer backend

### LINKED ISSUE #227: Support hosted signer backends (KMS / HSM / external signing service) in addition to in-memory keys
## Summary

Introduce a backend-agnostic `Signer` abstraction in `crates/server` so that signing material can live outside the Guardian process — in AWS KMS, GCP KMS, HashiCorp Vault Transit, a PKCS#11 HSM, or a custom HTTP signing service — instead of being loaded into memory from a file or from Secrets Manager. The current "load private key bytes into the process" path stays as the default for local dev and small deployments; hosted backends become opt-in for production.

## Motivation

Today the Guardian server holds its signing keys (notably the ack signer, plus any operator-side test fixtures used by `examples/demo`) in-process as raw key bytes sourced from one of two places:

- a file referenced by an env var (local / dev), or
- AWS Secrets Manager pulled at boot and decoded into memory (production deploy via `scripts/aws-deploy.sh`).

Both modes have the same blast radius: anything that reads process memory — a coredump, an `LD_PRELOAD` shim on the host, a debugger, an unintended `tracing::error!("{:?}", state)`, or a future supply-chain compromise of any dep — can extract the signing key. Once extracted, the key is permanently compromised and cosigners / downstream consumers must rotate trust.

Hosted signing backends (KMS, HSM, Transit, custom signer service) keep the private key outside the process boundary entirely:

- The Guardian server holds only a *key handle* (e.g. an AWS KMS ARN or a key ID).
- Signature requests cross a trust boundary (KMS API call, HSM command, HTTP request).
- Compromise of the running Guardian server lets an attacker *use* the key while the compromise is active, but does not exfiltrate it. After remediation, no key rotation is required and downstream trust is preserved.
- Audit becomes meaningfully better: KMS / Vault produce server-side audit logs of every signing call; today our only audit is the server's own logs.

This is the standard "use the right tool" answer for production deployments and aligns with how Falcon / ECDSA keys are handled in every comparable system (AWS Wallet, Coinbase Custody, gnosis-safe operator tooling, etc.).

A related issue https://github.com/OpenZeppelin/guardian/issues/226 narrows the blast radius for secrets that *must* live in memory. This issue removes the signer key from that set entirely for production deployments.

zeljkoX:
## New Features
* Added AWS KMS backend for ECDSA ACK signing; operators select backends via `GUARDIAN_ACK_ECDSA_BACKEND` environment variable (defaults to in-memory)
* New bootstrap command (`bootstrap-kms-ecdsa-key`) for KMS key provisioning

## Documentation
* Configuration reference for ECDSA backend selection and KMS key setup
* Production deployment procedures with KMS-backed ECDSA signer guidance
* Operator runbooks for secrets management with AWS KMS
* New task-oriented guides for AWS-managed ACK signers with Docker Compose examples


haseebrabbani (approved):
LGTM ✅

minor comment: KMS `Sign` is now on the hot path for every ECDSA ACK, so we're taking on KMS throttling/availability as a request-path dependency — a sign-failure/latency metric would surface any issues (which we can address with the coming Prometheus integration)

Issue #249, opened 2026-05-22
Dashboard & Observability


Dominik1999:
Tracking issue for operator observability — metrics, monitoring, API documentation, and human-readable feedback that give operators visibility into Guardian state.

**Parent:** [DRAFT] Guardian Management (#248)

## Related issues

- [x] #178 — Add operator dashboard metrics endpoints
- [x] #225 — Add native Prometheus integration
- [x] #99 — Add metrics and alarms for CloudWatch logs
- [x] #179 — Show human-readable error messages for wallet users
- [x] #241 — Add utoipa + model annotations to generate OpenAPI spec

Dominik1999:
In the future: Write logs to files + Datadog ...

PR #283, opened 2026-06-11
feat(server): add native Prometheus metrics endpoint

### LINKED ISSUE #225: Add native Prometheus integration to the Guardian
## Summary

Add native Prometheus integration to the Guardian server so operators can chart, alert on, and analyze service behavior over time.

The existing operator dashboard (`/dashboard/info`, `/dashboard/deltas`, `/dashboard/proposals`) exposes point-in-time snapshots intended for human operators. Prometheus complements this with continuous time-series collection suitable for Grafana dashboards, alerting, capacity planning, SLO tracking, and multi-instance aggregation.

This feature is additive and does not replace the existing dashboard APIs.

---

## Motivation

Today, the primary structured visibility into a running Guardian instance comes from:

* `tracing` log lines (useful for incident forensics, weak for long-term trends and alerting)
* `GET /dashboard/info` introduced by `005-operator-dashboard-metrics` (snapshot aggregates only; clients must poll repeatedly to derive time series)
* `GET /dashboard/deltas` and `/dashboard/proposals` (cursor-paginated resource listings, not operational aggregates)

This is sufficient for an operator UI, but not for operational observability use-cases such as:

1. **Trend analysis** — comparing canonicalization latency across time windows
2. **Alerting** — paging when proposal queues stall or in-flight work exceeds thresholds
3. **Capacity planning** — request throughput, P95/P99 latency, storage pool saturation
4. **Multi-instance aggregation** — federated metrics across multiple Guardian replicas
5. **SLO monitoring** — request success rate and latency objectives

Prometheus's pull-based scrape model with counters, gauges, and histograms is the de-facto standard for exposing operational telemetry, and integrates cleanly with existing observability stacks.

---

## Goals

* Expose a `/metrics` endpoint in Prometheus text exposition format
* Support standard Prometheus scraping without custom adapters
* Cover:

  * HTTP request path
  * gRPC request path
  * Storage backend operations
  * Canonicalization jobs
  * Multisig proposal lifecycle
  * Operator authentication/rate limiting
* Keep label cardinality strictly bounded
* Keep existing `/dashboard/*` endpoints unchanged
* Ensure metrics scraping remains lightweight and independent from live database queries

---

## Non-goals

* Shipping Grafana dashboards or alert rules in this issue
* Long-term retention or remote-write configuration
* Distributed tracing / OpenTelemetry integration
* Exemplars or trace correlation
* Authenticated multi-tenant metrics access
* Replacing `/dashboard/info`
* gRPC reflection of metrics

v1 assumes the metrics endpoint is either:

* reachable only on a private network, or
* protected by a reverse proxy / sidecar

A simple optional shared-secret bearer token is sufficient for initial deployments.

---

---

## Acceptance criteria

* [ ] `GET /metrics` returns valid Prometheus exposition format
* [ ] Metrics listener respects configured bind address and path
* [ ] Core metric taxonomy is populated under realistic load
* [ ] Label sets respect cardinality rules
* [ ] No high-cardinality identifiers appear in labels
* [ ] Slow aggregate metrics are refreshed asynchronously, not during scrape
* [ ] Scrapes do not perform live database scans
* [ ] `GUARDIAN_METRICS_ENABLED=false` disables:

  * metrics listener
  * recorder/exporter setup
  * background refresh tasks
* [ ] Bearer token protection returns `401` for missing/invalid tokens
* [ ] Unit tests cover:

  * route template normalization
  * label allowlist enforcement
  * refresher task ticking
  * bearer-token guard
* [ ] `examples/operator-smoke-web` mentions the metrics endpoint where relevant
* [ ] `smoke-test-operator-dashboard` references metrics scraping where appropriate
* [ ] `spec/api.md` and `spec/components.md` document the metrics surface and configuration
* [ ] Local Prometheus scraping against a running `guardian-server` produces non-zero HTTP and storage metrics after smoke flows

---

## Out of scope (follow-ups)

* Grafana dashboards
* PromQL alert rules and runbooks
* Distributed tracing / OpenTelemetry
* Long-term retention configuration
* Remote-write integration
* Exemplars / trace correlation
* Replacing `/dashboard/info`

Potential future follow-up:

* optional local observability stack (`prometheus` + `grafana`) via docker-compose for development and smoke testing

---
* Prometheus best practices:

  * https://prometheus.io/docs/practices/naming/
  * https://prometheus.io/docs/practices/instrumentation/

haseebrabbani:
Closes #225 (part of #249 Dashboard & Observability)

## Summary

Adds an opt-in native Prometheus integration to `guardian-server`: a `/metrics` endpoint in the classic text exposition format, served from a dedicated listener, covering the HTTP and gRPC request paths, storage backend operations, canonicalization jobs, the multisig proposal lifecycle, operator auth, and rate limiting. Additive only — no `/dashboard/*` behavior changes.

## Design

- **Library**: `metrics` facade + `metrics-exporter-prometheus` (+ `metrics-process` for standard `process_*` metrics). Chosen over the official `prometheus-client` to instrument deep call sites without threading a registry through `AppState`; both render identically to Prometheus.
- **Dedicated listener**: `GUARDIAN_METRICS_ADDR` (default `127.0.0.1:9464`), path configurable. The main API port returns 404 for `/metrics`; the metrics listener bypasses rate limiting/CORS.
- **Protection** (layered per the Prometheus security model): loopback bind by default → optional shared-secret bearer token (`401` otherwise, constant-time compare via the existing `subtle`-based helper, token held as `SecretString`) → TLS delegated to a reverse proxy / private network.
- **Scrapes never hit storage**: slow cross-account aggregates are computed by a background refresher (default 30s) and published as gauges, with a staleness timestamp gauge and failure counter.
- **Bounded cardinality**: route labels use axum templates (`/dashboard/accounts/{account_id}`), unmatched paths collapse into one series, gRPC labels come from the proto, and everything else is a closed enum. The taxonomy lives in `crates/server/src/metrics/names.rs` as a static registry; a unit test structurally enforces the label-key allowlist and naming conventions.
- **`GUARDIAN_METRICS_ENABLED=false` (default)** disables the listener, recorder, storage decorator, and refresher entirely.

## New configuration

| Variable | Default |
|---|---|
| `GUARDIAN_METRICS_ENABLED` | `false` |
| `GUARDIAN_METRICS_ADDR` | `127.0.0.1:9464` |
| `GUARDIAN_METRICS_PATH` | `/metrics` |
| `GUARDIAN_METRICS_REFRESH_INTERVAL_SECS` | `30` |
| `GUARDIAN_METRICS_BEARER_TOKEN` | unset (open; rely on network isolation) |

## Testing

- 37 new tests: route-template normalization, label-allowlist enforcement, refresher ticking (paused tokio time), bearer-guard matrix, gRPC trailer/header/cancel status capture, storage decorator outcomes, plus an end-to-end exposition integration test.
- Full suite green (590 passed), `cargo fmt` clean, `clippy --all-targets --all-features -- -D warnings` clean.
- Verified live: 401 without/with wrong token, 200 + exposition with token, templated route labels with no raw identifiers, real gRPC call recorded with status from HTTP/2 trailers, gauges populated by the refresher, connection refused on 9464 when disabled.

## Summary by CodeRabbit

* **New Features**
  * Added optional Prometheus metrics exposition endpoint with configurable bearer token authentication.
  * Integrated observability across HTTP/gRPC requests, storage operations, and business logic events.
  * Metrics endpoint runs on dedicated listener, configurable via environment variables.
  * Background refresh of cross-account aggregate gauges.

* **Documentation**
  * Updated configuration and production guides with metrics setup instructions.
  * Added metrics taxonomy to API specification.


crates/server/src/services/sign_delta_proposal.rs
```diff
@@ -160,6 +160,11 @@ pub async fn sign_delta_proposal(
         .update_delta_proposal(&normalized_commitment, &delta_proposal)
         .await
         .map_err(GuardianError::StorageError)?;
+    metrics::counter!(
+        crate::metrics::names::PROPOSALS_TOTAL,
+        crate::metrics::names::LABEL_EVENT => "signed"
```
zeljkoX: Let's centralize label values behind enums.

Similar to crate::metrics::names, we can expose crate::metrics::labels and use.

crates/server/src/services/delta_commit.rs
```diff
@@ -124,18 +124,27 @@ impl DeltaCommitStrategy {
...
                         .await
                     {
-                        warn!(
-                            account_id = %delta.account_id,
-                            proposal_id = %id,
-                            error = %e,
-                            "Failed to delete proposal, but continuing"
-                        );
+                        Ok(()) => {
+                            metrics::counter!(
```
zeljkoX: We already have this in crates/server/src/jobs/canonicalization/processor.rs.

Potentially we would double count.

Here we can track number of deleted proposals for example

crates/server/src/services/push_delta.rs
```diff
@@ -132,6 +132,15 @@ pub async fn push_delta(state: &AppState, params: PushDeltaParams) -> Result<Pus
             &new_commitment,
         )
         .await?;
+    metrics::counter!(
+        crate::metrics::names::DELTAS_SUBMITTED_TOTAL,
+        crate::metrics::names::LABEL_KIND => if matching_proposal_payload.is_some() {
+            "proposal_commit"
+        } else {
+            "delta"
```
zeljkoX: maybe to use "direct" here, meaning it was submitted directly.

crates/server/src/metrics/names.rs
```diff
@@ -0,0 +1,429 @@
...
+        axum::http::Method::HEAD => "HEAD",
+        axum::http::Method::OPTIONS => "OPTIONS",
+        _ => "other",
+    }
+}
+
+/// Split a gRPC request path (`/package.Service/Method`) into bounded
+/// `(service, method)` label values. Paths that do not have the
+/// expected two-segment shape collapse into `("unknown", "unknown")`.
+pub fn normalize_grpc_method(path: &str) -> (&str, &str) {
```
zeljkoX: **unbounded gRPC label cardinality.** This returns the raw two path segments as `service`/`method` for any shape, and `MetricsGrpcLayer` wraps `Server::builder()` **before** tonic routing — so unrouted requests (which tonic answers with a trailers-only `unimplemented`) still reach here with caller-controlled path segments. A client spraying random `/x.Y/Z` paths can mint unbounded `{service,method}` series — a metric-cardinality DoS in a feature whose premise is bounded cardinality (the HTTP side already guards this via `MatchedPath → "unmatched"`). Suggest allowlisting the known proto service/method names (there's one service, `guardian.Guardian`) and collapsing anything else to `("unknown","unknown")`.

crates/server/src/metrics/names.rs
```diff
@@ -0,0 +1,429 @@
...
+        name: DELTAS_SUBMITTED_TOTAL,
+        kind: MetricKind::Counter,
+        labels: &[LABEL_KIND],
+        help: "Deltas accepted by the server, by submission kind (delta, proposal_commit).",
+    },
+    MetricDef {
+        name: PROPOSALS_TOTAL,
+        kind: MetricKind::Counter,
+        labels: &[LABEL_EVENT],
+        help: "Multisig proposal lifecycle events (created, signed, threshold_reached, \
```
zeljkoX: **Help text advertises an unemitted value.** No call site emits `threshold_reached` (grep-verified) and `spec/api.md` lists only `created/signed/finalized`. Either drop it here + in the spec, or actually emit it in `sign_delta_proposal` when the signature count first crosses threshold — that's the real `ProposalStatus::Ready` transition. Either way keep help/spec/impl in lockstep, ideally by anchoring the value set in a `ProposalEvent` enum so it can't drift again.

crates/server/src/metrics/recorder.rs
```diff
@@ -0,0 +1,104 @@
...
+const DURATION_BUCKETS: &[f64] = &[
+    0.001, 0.0025, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1.0, 2.5, 5.0, 10.0,
+];
+
+/// Build an uninstalled recorder. The caller decides whether to
+/// install it globally (production) or scope it locally (tests).
+pub fn build_recorder() -> PrometheusRecorder {
+    PrometheusBuilder::new()
+        .set_buckets_for_metric(
+            Matcher::Suffix(DURATION_SECONDS_SUFFIX.to_string()),
```
zeljkoX, line 40: **one bucket set is wrong for the canonicalization-run histogram.** This `Suffix("duration_seconds")` rule applies the 1ms–10s buckets to *every* `*_duration_seconds` metric, including `guardian_canonicalization_run_duration_seconds` — a full pass over all accounts, which routinely exceeds 10s and saturates `+Inf`, making the histogram useless for its own workload. Add a second `set_buckets_for_metric` rule (by full name) for the canonicalization-run histogram extending to ~30/60/120/300s. (The openzeppelin-relayer metrics use per-metric buckets for exactly this reason.)

crates/server/src/builder/handle.rs
```diff
@@ -257,6 +314,7 @@ impl ServerHandle {
                 tracing::info!(address = %addr, "gRPC server listening");
 
                 Server::builder()
+                    .layer(MetricsGrpcLayer::new(metrics_enabled))
```
zeljkoX: **disabled path isn't free on gRPC.** This attaches `MetricsGrpcLayer` unconditionally; with `enabled=false` the layer still `Box::pin`s the future and wraps every response body. The HTTP side attaches `track_http` only `if metrics_enabled` — match that here (attach the gRPC layer only when enabled, or short-circuit `call()` to the inner service without wrapping) so "disabled → nothing runs" holds for both transports.

crates/server/src/builder/handle.rs
```diff
@@ -59,6 +64,50 @@ impl ServerHandle {
...
+        // requests are never served by an uninstrumented stack. The
+        // global recorder is process-wide and set-once: tolerate an
+        // already-installed recorder (repeated run() in one process,
+        // e.g. e2e harnesses) instead of panicking. Tests never
+        // install globally — they scope local recorders.
+        let metrics_enabled = self.metrics_config.enabled;
+        if metrics_enabled {
+            let recorder = crate::metrics::build_recorder();
+            let handle = recorder.handle();
+            if let Err(error) = metrics::set_global_recorder(recorder) {
```
zeljkoX: **Fix or document — orphan handle on repeated `run()`.** If `set_global_recorder` fails (already installed), execution continues but `handle` belongs to the *new, uninstalled* recorder, so the metrics router renders empty while instrumentation writes to the pre-existing global recorder. Harmless in prod (single `run()`), but it bites repeated-`run()`/e2e harnesses. Either fail metrics startup, or take the handle from the already-installed recorder; at minimum comment that the served handle won't reflect the existing recorder.

crates/server/src/jobs/canonicalization/processor.rs
```diff
@@ -356,19 +372,29 @@ impl DeltasProcessorBase {
...
-                tracing::warn!(
-                    account_id = %delta.account_id,
-                    proposal_id = %id,
-                    error = %e,
-                    "Failed to delete proposal, but continuing"
-                );
+                Ok(()) => {
+                    metrics::counter!(
+                        crate::metrics::names::PROPOSALS_TOTAL,
+                        crate::metrics::names::LABEL_EVENT => "finalized"
```
zeljkoX: **Same `finalized` placement bug as `delta_commit.rs` — fix in both files.** Gated on delete success in the `Ok(())` arm; **move the increment up — out of the `Ok(())` arm to just inside the `if let`, above the delete `match`** — so finalization is counted regardless of cleanup outcome. The identical move is required in `delta_commit.rs` (Optimistic mode); the two blocks are structurally identical.

This is the **Candidate-mode** path (canonicalization worker); the Optimistic emit lives in `delta_commit.rs`. They are mutually exclusive per deployment (`DeltaCommitStrategy::from_app_state`), so *not* a double-count — keep both. Note this path won't self-heal a missed count: once the delta is canonical it is no longer a candidate, so the worker never revisits it to retry the delete.

crates/server/src/services/push_delta.rs
```diff
@@ -132,6 +132,15 @@ pub async fn push_delta(state: &AppState, params: PushDeltaParams) -> Result<Pus
             &new_commitment,
         )
         .await?;
+    metrics::counter!(
+        crate::metrics::names::DELTAS_SUBMITTED_TOTAL,
+        crate::metrics::names::LABEL_KIND => if matching_proposal_payload.is_some() {
```
zeljkoX: **Refine label.** (1) `matching_proposal_payload` returns `None` both when there is no proposal *and* when the lookup errored (it swallows storage errors to `None`), so under storage faults a proposal-commit is mislabeled `"delta"` — distinguish or document the error case. (2) `kind="delta"` is tautological on `deltas_submitted_total`; prefer `"direct"` vs `"proposal_commit"` (new metric, no back-compat concern). (3) Name says "submitted" but the help says "accepted" — pick one. Good candidate for a `DeltaKind { Direct, ProposalCommit }` enum to kill the stringly-typed literal.

crates/server/src/services/delta_commit.rs
```diff
@@ -124,18 +124,27 @@ impl DeltaCommitStrategy {
...
+                                crate::metrics::names::PROPOSALS_TOTAL,
+                                crate::metrics::names::LABEL_EVENT => "finalized"
```
zeljkoX: **`finalized` is gated on delete *success*, mislabeling the event.** The proposal is finalized the moment we are inside the `if let` (delta canonical + matching proposal found); the delete is just cleanup. Counting only in the `Ok(())` arm undercounts finalizations whenever the cleanup delete errors. **Move the increment up — out of the `Ok(())` arm to just inside the `if let`, above the `match`.**

**The identical move is required in `processor.rs` (Candidate-mode path) as well** — the two blocks are structurally identical; fixing only one leaves that mode undercounting on delete failure.

Not a double-count: this **Optimistic-mode** emit and the `processor.rs` one (**Candidate mode**) are mutually exclusive per deployment (`DeltaCommitStrategy::from_app_state`), so both are required — keep both. Cleanup-delete failures remain observable via `storage_operations_total{operation="delete_delta_proposal",outcome="error"}`, so no separate deleted-proposals counter is needed.

zeljkoX (commented):
Looks great! 

Thanks for hard work!

Few things that would be beneficial to operators:
- Miden RPC outbound metrics: guardian_miden_rpc_requests_total{operation,outcome} and guardian_miden_rpc_duration_seconds{operation} with a closed operation set. The upstream chain node is currently the most important SLI with zero instrumentation (network/miden/ has no metrics:: calls).
- DB pool saturation metrics from deadpool status(): current/max/available/waiting gauges, and ideally an acquire-wait histogram. Pool exhaustion is a silent latency cliff today.
- Account lifecycle counters, especially guardian_accounts_created_total (the current guardian_accounts gauge can't answer "created in last N min" — rate()/increase() need a counter). Keep the gauge for point-in-time total.
- gRPC in-flight gauge for parity with HTTP.


To complete the picture, we should add a Grafana example to the docs/guides. Since this PR is already self-contained and large, could you create an issue and tackle it next?

haseebrabbani:
Thanks for the review! Addressed all feedback:

**New instrumentation from your summary:**

- **Miden RPC outbound metrics** — added `guardian_miden_rpc_requests_total{operation,outcome}` and `guardian_miden_rpc_duration_seconds{operation}`. The only outbound chain-node call today is `get_account_commitment` (in `verify_state`), so the closed operation set starts there and grows as RPC call sites are added.
- **DB pool saturation** — added a `pool_status()` hook on `StorageBackend` (Postgres maps deadpool's `status()`; filesystem returns `None`), published by the refresher as `guardian_db_pool_connections_max` / `_connections` / `_connections_available` / `_pending_acquires`. One deliberate omission: the acquire-wait histogram — deadpool doesn't expose acquire hooks to instrument, and acquire wait is already folded into `guardian_storage_operation_duration_seconds`, so exhaustion shows up there plus in `_pending_acquires`. Can revisit if we ever wrap `pool.get()` directly.
- **`guardian_accounts_created_total{kind="miden"|"evm"}`** — emitted on first-time creation only (`/configure` and `/evm/accounts` both also serve re-configuration); the point-in-time gauge stays.
- **gRPC in-flight gauge** — `guardian_grpc_requests_in_flight`, held by a drop guard so client cancellations and transport errors release the slot.

**Inline comments:**

- **Orphan recorder handle** — metrics wiring (refresher + listener) is now gated on `set_global_recorder` succeeding; on failure it logs and skips, so `/metrics` can never serve an empty registry while instrumentation writes to the pre-existing recorder.
- **gRPC label cardinality** — `normalize_grpc_method` now matches against a `KNOWN_GRPC_METHODS` allowlist (the 11 Guardian methods + reflection); anything else collapses into `("unknown","unknown")`. Added a test that parses `proto/guardian.proto` and fails if the allowlist drifts.
- **Label enums** — new `metrics/labels.rs` with `Outcome`, `ProposalEvent`, `DeltaKind`, `CandidateOutcome`, `AccountKind`; all call sites migrated off string literals.
- **`threshold_reached`** — dropped from help text and spec rather than emitted: the server has no threshold-crossing transition to hook (signatures accumulate on `Pending` proposals; the threshold is only evaluated at commit), so emitting it would invent a state the server doesn't track.
- **`finalized` placement** — moved above the cleanup delete in both `delta_commit.rs` (optimistic) and `processor.rs` (candidate), with comments noting the two emits are mutually exclusive per deployment. Delete failures stay visible via `storage_operations_total{operation="delete_delta_proposal",outcome="error"}`, so no separate deleted-proposals counter.
- **`kind="delta"` → `"direct"`** — renamed via the enum; help text verb aligned; the lookup-swallows-errors caveat is documented at the call site.
- **Canonicalization run buckets** — added a `Matcher::Full` override with buckets up to 300s (full-name matchers take precedence over the suffix rule).
- **Disabled gRPC path** — the metrics layer is now attached only when metrics are enabled, matching the HTTP side; disabled means no wrapping at all.

**Grafana example** — agreed on keeping it out of this PR; opened #285 to track it and I'll pick it up next.

crates/server/src/metrics/storage.rs
```diff
@@ -66,6 +66,10 @@ impl StorageBackend for InstrumentedStorage {
         self.inner.kind()
     }
 
+    fn pool_status(&self) -> Option<crate::storage::PoolStatus> {
```
zeljkoX, line 69: Storage has 2 pools,

- storage (`PostgresService`) | `GUARDIAN_DB_POOL_MAX_SIZE` (default 16) | delta/state storage, canonicalization |
- metadata (`PostgresMetadataStore`) | `GUARDIAN_METADATA_DB_POOL_MAX_SIZE` (defaults to storage size) | account metadata, dashboard listings, operator auth, **and audit** (auditor shares it via `metadata.pool_handle()`, line 120) |

So `guardian_db_pool_*` reflects the storage pool only. The metadata pool — which backs all the *interactive* paths (auth, dashboard, audit) and has its own independently-tunable max size — is unmonitored. 

Let's expose metrics for second pool as well.

zeljkoX (approved):
Added one minor comment. 

Other than that looks great! Thanks

crates/server/src/metrics/storage.rs
haseebrabbani: Good catch — added a `pool_status()` hook to `MetadataStore` (mirroring the storage one) and a `pool` label (`storage` / `metadata`) on the `guardian_db_pool_*` gauges, so both pools are now exposed:

​```
guardian_db_pool_pending_acquires{pool="storage"}   5
guardian_db_pool_pending_acquires{pool="metadata"}  0
​```

Used a label rather than separate metric names so PromQL can aggregate or split as needed. Note the two pools can report different `_connections_max` —  `GUARDIAN_METADATA_DB_POOL_MAX_SIZE` is independently tunable (defaults to the storage size) — so a mismatch there is expected, not a bug

PR #292, opened 2026-06-18
docs(speckit): human-readable wallet error messages spec (#179)

haseebrabbani:
## What

Draft **spec only** (no implementation yet) for issue #179 — surface short, user-friendly error messages for common wallet failures instead of raw backend/transport errors.

Adds `speckit/features/009-human-readable-errors/spec.md` following the house spec-kit format.

Refs #179.

## Approach (proposed)

Server-side single source of truth:

- Add `GuardianError::user_message()` — a short, **sanitized** sentence per variant (no account IDs / commitments / paths / raw upstream text).
- Surface it **additively** as `user_message` on the HTTP error envelope **and** the gRPC `Status` details (generalizing the existing `GUARDIAN_ACCOUNT_PAUSED` details pattern). `code`, `error`, and all status mappings stay frozen.
- Rust + TS clients consume it; only **codeless transport failures** (server never reached) are classified client-side — the pattern from 0xMiden/wallet #273 (`connectivity-classify.ts`).
- i18n stays client-side keyed on the stable `code`; server ships English defaults.

Anchored to the constitution: HTTP/gRPC parity (II), bottom-up propagation to Rust+TS clients (I), additive/stable boundary errors (IV). Bonus: stops the current raw `error` strings (which embed IDs/commitments/raw RPC text) from leaking to wallet clients.

## Open questions for reviewers (@MCarlomagno @zeljkoX)

1. Final wire field name — `user_message` vs `hint` vs `display_message`.
2. Granular per-cause `code` for internal errors, or collapse to one `internal_error`.
3. Ship the starter copy catalog now, or gate wording on UX review.
4. Emit on all surfaces (incl. operator dashboard) vs wallet-only.

The 2026-06-18 clarifications in the spec are marked **proposed — pending confirmation**.

## Not included

Implementation, `plan.md` / `data-model.md` / `contracts/`. Those follow once direction is confirmed (per `AGENTS.md` §4 contract-change workflow: proto + OpenAPI regen → Rust client → TS clients → multisig SDK → examples).

## Summary by CodeRabbit

* **Documentation**
  * Added a feature specification for clearer, user-safe error messages across wallet-facing APIs.
  * Defines a consistent error shape with stable codes and readable messages for both HTTP and gRPC responses.
  * Documents how common connectivity and internal failures should be presented, including behavior for unknown future error codes.

### DIFF
--- a/speckit/features/009-human-readable-errors/spec.md
+++ b/speckit/features/009-human-readable-errors/spec.md
@@ -7,30 +7,46 @@
 
 ## Problem
 
-Guardian already emits a stable, machine-readable error `code` for every failure (`GuardianError::code()` in `crates/server/src/error.rs`), and the operator client already documents "branch on `code`, not the human message" (`packages/guardian-operator-client/src/types.ts`). What it does **not** provide is a short message safe to show directly to a wallet end-user:
+Guardian already emits a stable, machine-readable error `code` for every failure (`GuardianError::code()` in `crates/server/src/error.rs`), and the operator client already documents "branch on `code`, not the human message" (`packages/guardian-operator-client/src/types.ts`). What it does **not** provide is a message safe to show directly to a wallet end-user:
 
-- The HTTP envelope's `error` field and the gRPC `Status` message are the `Display` string — developer/diagnostic text that embeds account IDs, commitments, nonces, and raw upstream/RPC error text (e.g. `"Commitment mismatch: expected 0xaa, got 0xbb"`, `"RPC unavailable: <raw transport error>"`, `"Insufficient signatures: required 3, got 1"`).
+- The HTTP envelope's `error` field and the gRPC `Status` message are the `Display` string — developer/diagnostic text that embeds account IDs, commitments, nonces, and raw upstream/RPC error text (e.g. `"Commitment mismatch: expected 0xaa, got 0xbb"`, `"RPC unavailable: <raw transport error>"`). Handing that to an untrusted wallet is both unfriendly and a disclosure risk.
 - The wallet-facing clients (`packages/miden-multisig-client`, `crates/client`) perform no error translation — they rethrow raw gRPC/transport errors.
 
-Net effect: a wallet user sees raw backend or transport noise. This feature adds a curated, end-user-safe message layer with the server as the single source of truth.
+Net effect: a wallet user sees raw backend or transport noise. This feature reshapes the error wire contract around a single user-safe `message`, with the server as the source of truth, and keeps the diagnostic `Display` text off the wire entirely (server logs only).
 
 ## Clarifications
 
-### Session 2026-06-18 (proposed resolutions — pending confirmation with @MCarlomagno / @zeljkoX)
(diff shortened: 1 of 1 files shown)


zeljkoX (commented):
Since Guardian is still early-phase with no external consumers locked to the current contract, I think we should treat this as a breaking change rather than add user_message additively. The additive path keeps two human-ish strings whose names don't say which is safe to show, plus a flat envelope that grows an optional field per variant. A clean reshape is cheap now.

Proposed wire shape — one object, identical on the HTTP body and the gRPC Status.details:

```
{
  "code": "authorization_failed",
  "message": "You're not an authorized signer for this account.",
  "meta": { "retryable": false }
}
```

- code — stable machine code (existing vocabulary, unchanged). Branch key + i18n key.
- message — user-safe sentence, always present and safe to display. Only code is stable; wording can change.
- meta — structured machine data; holds the fields that are top-level optionals today (retryable, retry_after_secs, missing_permissions, paused_at, paused_reason). Omitted when empty.

The diagnostic string stays off the wire — server logs only. The Display text (account IDs, commitments, postgres://…, raw RPC errors) is exactly what we don't want to hand an untrusted wallet. Logging it instead of returning it removes the disclosure risk by construction. This also frees us to name the field plainly message (the old diagnostic error field is gone). We also drop success (the HTTP status is the discriminator; gRPC never had it).

Examples:

```
{ "code": "rate_limit_exceeded", "message": "Too many requests — please try again shortly.", "meta": { "retryable": true, "retry_after_secs": 30 } }
{ "code": "storage_error", "message": "Something went wrong on Guardian's side. Please try again.", "meta": { "retryable": true } }
```

gRPC: Status.code unchanged, Status.message becomes the user-safe message, and Status.details carries the same {code, message, meta} object. Details is still needed because the ~16 canonical gRPC codes can't recover our 37 stable code values (e.g. authorization_failed / signer_not_authorized both map to PERMISSION_DENIED).

haseebrabbani:
@zeljkoX Thanks for the feedback, pushed an update to reshape the spec accordingly:

- Single `{ code, message, meta }` object, byte-identical on the HTTP body and gRPC `Status.details`; `Status.message` = the user-safe `message`.
- Diagnostic `Display` text is off the wire entirely (server logs only), and `success`/`error` are gone. The `meta` block absorbs `retryable`, `retry_after_secs`, `missing_permissions`, `paused_at`, `paused_reason`, omitted when empty.
- `code` vocabulary and HTTP/gRPC status mappings unchanged; framed as a coordinated breaking change across server → proto/OpenAPI → Rust client → TS clients (incl. operator-client) → examples.
- Also folded in Copilot's catch: explicit split between server-mapped connectivity faults (`network_error` / `rpc_unavailable` / `rpc_validation_failed`) and pure internal faults, so the requirements/catalog no longer contradict.

Three things still open in the spec if you have a preference:

1. **OQ-2** — keep granular internal codes sharing one message vs collapse to `internal_error` (I'm leaning keep-granular).
2. **OQ-3** — ship the copy catalog now vs gate wording on UX review.
3. **OQ-5** — populate `meta.retryable` for every variant vs only where non-obvious.

zeljkoX:
LGTM

PR #293, opened 2026-06-19
feat: Storage encryption support

### LINKED ISSUE #20: Storage encryption
Storage layer is plaintext, we want to setup an encryption key in the PSM. So the contents of the account state and deltas are stored encrypted and decrypted in transit 

in private mode: the enclave holds the encryption keys for operating over contents in-transit

zeljkoX:
This PR adds an application-layer confidentiality boundary for Guardian's sensitive storage blobs (account state JSON, delta payloads, proposal payloads). Payloads are authenticated-encrypted before being written and transparently decrypted on read, using a 32-byte key that lives outside the database.

 How it works

  - A StorageDecorator wraps both filesystem and Postgres backends, encrypting sensitive fields while leaving routing/index columns (account
  id, nonce, lifecycle status) plaintext.
  - AEAD cipher with an envelope format; ciphertext is bound to record identity to prevent cross-record swaps.
  - Pluggable key provider: raw key (GUARDIAN_STORAGE_ENCRYPTION_KEY) or KMS/Secrets Manager (GUARDIAN_STORAGE_ENCRYPTION_KEY_SECRET_ID /
  _KEY_ID).
  - A marker mechanism distinguishes encrypted vs. plaintext records so upper layers are unchanged.

## Release Notes

* **New Features**
  * Added opt-in storage encryption at rest for sensitive payloads using AES-256-GCM
  * AWS Secrets Manager integration for managing encryption keys in production
  * Bootstrap command to initialize encryption key infrastructure

* **Configuration**
  * New environment variables for encryption configuration: `GUARDIAN_STORAGE_ENCRYPTION_KEY`, `GUARDIAN_STORAGE_ENCRYPTION_KEY_ID`, and `GUARDIAN_STORAGE_ENCRYPTION_SECRET_NAME`

* **Bug Fixes**
  * Storage retrieval errors now propagate correctly instead of silently returning empty results


haseebrabbani (approved):
lgtm, minor comments:

- **Bootstrap key via argv.** nitpick but `create-secret --secret-string "$secret_value"` exposes the key in the process arg list (`ps`/`/proc`). Prefer `--secret-string file://<restricted-temp>`.

- **Doc the confidentiality boundary.** Only payload JSON is encrypted; routing/index columns (account_id, nonce, status, commitment) stay plaintext by design. Worth stating explicitly in the threat-model docs so operators don't over-assume what "encrypted at rest" covers here.

PR #301, opened 2026-06-24
feat: scalability improvements

### LINKED ISSUE #190: Decouple Guardian canonicalization from API instances
## Summary
Guardian’s current canonicalization flow runs as an in-process background worker inside the server. That works for a single instance, but it becomes a bottleneck and a correctness risk when scaling Guardian horizontally across multiple machines.

We should separate canonicalization from the main request-serving path so multiple Guardian API instances can accept traffic, while a dedicated canonicalization worker is responsible for finalizing candidate deltas.

## Problem
Today, canonicalization is driven by per-instance polling and account metadata flags. This creates a few scaling and coordination problems:
- API instances and canonicalization logic are tightly coupled.
- Multi-instance deployments risk duplicate processing and race conditions.
- Candidate creation and canonicalization scheduling are not modeled as a durable shared job flow.
- Finalization touches multiple pieces of state and should be coordinated more explicitly.

## Goal
Introduce a high-level architecture where:
- Guardian API instances only validate and persist candidate deltas.
- Candidate creation also creates a durable canonicalization task in shared storage.
- A separate canonicalization worker/service consumes those tasks and performs finalization.
- Canonicalization remains sequential per account and safe across restarts.
- The system can support multiple Guardian API instances without relying on in-process background jobs.

## Expected outcome
After this change, we should be able to run:
- many Guardian API nodes for request handling
- one dedicated canonicalization worker initially
- a path to safely support multiple workers later through proper claiming/locking semantics

## Acceptance criteria
- Canonicalization execution is no longer tied to every Guardian API process.
- Candidate persistence and canonicalization task creation are coordinated durably.
- A worker can recover unfinished canonicalization tasks after restart.
- Finalization updates are applied consistently and do not leave partially completed state.
- Per-account ordering is preserved during canonicalization.
- Failed canonicalizations end in an explicit terminal or retryable state.
- The design is validated for multi-instance deployment with Postgres/shared storage.

### LINKED ISSUE #242: Ensure horizontal scaling works correctly across multiple Guardian instances
**Context:**
The prod stage profile scales the guardian-server ECS service between 2 and 6 tasks based on CPU/memory. The ALB distributes requests round-robin. Several pieces of in-memory state and background work were written under an implicit single-instance assumption. 

**Ensure horizontal scaling works correctly across multiple Guardian instances:**

 - Canonicalization worker runs on every instance with no leader election #190
 - Operator/EVM sessions and auth challenges are per-process
- Rate limiter is per-process
- Dashboard cursor secret silently breaks pagination across instances if not set
-  Filesystem backend: refuse to start in prod stage; document as dev-only.

**Acceptance criteria**

- Operator can complete the full login flow successfully with 2+ replicas behind the ALB.
- A delta is canonicalized exactly once regardless of replica count.
- Rate limits enforced consistently across replicas (within some documented tolerance).
- Pagination cursors valid across all replicas.
- Operator runbook documents which environment variables must be set for HA and any state-store dependencies.

zeljkoX:
**Horizontal scaling correctness across replicas**

Makes the Guardian server correct under 2–6 replicas behind a non-sticky load balancer. Several subsystems were written for a single instance, so a request that started on one replica and continued on another could fail, and background work ran redundantly on every replica.

What changed
- Auth — operator/EVM challenges and sessions move from per-process memory to shared storage, so login (challenge → verify) and authenticated requests work on any replica; logout is honored fleet-wide.
- Canonicalization — now runs on exactly one replica via a leader lease with a fencing token, concurrent renewal, and cooperative cancellation, so each candidate is promoted/discarded once (no double-processing or retry double-counts).
- Rate limiting — partitioned by GUARDIAN_MAX_REPLICAS (wired from the autoscaling max capacity) so aggregate enforcement stays at/below the global limit instead of multiplying per replica.
- Guardrails & ops — prod refuses the filesystem backend, migrations are serialized across simultaneous boots, a single startup log line reports the resolved coordination mode, plus an HA runbook and a local docker-compose guide.

> Design note: the goal was explicitly not to introduce Redis (or any new infrastructure dependency). All coordination — sessions, challenges, and the canonicalization lease — reuses the existing Postgres backend that the prod image already ships with. The deployment surface is unchanged; coordination mode is derived from the storage backend alone (Postgres = shared, filesystem = in-memory single-process), so dev/local needs no new services.

## Summary by CodeRabbit

* **New Features**
  * Added horizontal-scaling support with shared authentication sessions and login challenges across replicas.
  * Added coordinated worker leadership and fencing to prevent duplicate or stale canonicalization writes.
  * Added configurable replica-aware rate limiting through `GUARDIAN_MAX_REPLICAS`.
  * Added production safeguards, shared dashboard cursor secret support, and a local two-replica setup guide.

* **Bug Fixes**
  * Authentication and logout now fail closed when shared storage is unavailable.
  * Expired sessions and challenges are cleaned up automatically.
  * Improved migration safety and cross-replica session revocation consistency.

* **Documentation**
  * Added configuration, deployment, horizontal-scaling runbooks, guides, and updated API error documentation.


haseebrabbani:
Two comments after latest changes — everything else looks good:

**1. Default prod rate-limits operator login to ~2 challenges/min per commitment** — `dashboard/config.rs:67`
The per-commitment challenge limiter is divided by `GUARDIAN_MAX_REPLICAS`, which Terraform now defaults to the surge capacity (prod: `max(desired, 6) × 200% = 12`). That yields burst `5/12 → 1/sec` and sustained `30/12 = 2/min` per replica — an operator pinned to one replica by keep-alive gets a 429 on their third challenge within a minute. Unlike the global limits, these are compile-time constants with no env override, so there's no operator remedy. Suggest partitioning the login limiter by autoscaling max (not surge), or making the constants configurable.

**2. `kms:Decrypt` grant breaks on non-ARN key references** — `infra/iam.tf:63`
`data.aws_secretsmanager_secret.dashboard_cursor[0].kms_key_id` returns whatever identifier the secret was created with (bare key ID, `alias/...`, or ARN), but an IAM policy `Resource` must be an ARN — a pre-created secret using `--kms-key-id alias/...` (a documented path via `GUARDIAN_DASHBOARD_CURSOR_SECRET_NAME`) fails apply with `MalformedPolicyDocument`. The other prod secrets (ACK keys, storage encryption) carry no identity-policy KMS statement and delegate to key policy — suggest doing the same here.

zeljkoX:
> Two comments after latest changes — everything else looks good:
> 
> **1. Default prod rate-limits operator login to ~2 challenges/min per commitment** — `dashboard/config.rs:67` The per-commitment challenge limiter is divided by `GUARDIAN_MAX_REPLICAS`, which Terraform now defaults to the surge capacity (prod: `max(desired, 6) × 200% = 12`). That yields burst `5/12 → 1/sec` and sustained `30/12 = 2/min` per replica — an operator pinned to one replica by keep-alive gets a 429 on their third challenge within a minute. Unlike the global limits, these are compile-time constants with no env override, so there's no operator remedy. Suggest partitioning the login limiter by autoscaling max (not surge), or making the constants configurable.
> 
> **2. `kms:Decrypt` grant breaks on non-ARN key references** — `infra/iam.tf:63` `data.aws_secretsmanager_secret.dashboard_cursor[0].kms_key_id` returns whatever identifier the secret was created with (bare key ID, `alias/...`, or ARN), but an IAM policy `Resource` must be an ARN — a pre-created secret using `--kms-key-id alias/...` (a documented path via `GUARDIAN_DASHBOARD_CURSOR_SECRET_NAME`) fails apply with `MalformedPolicyDocument`. The other prod secrets (ACK keys, storage encryption) carry no identity-policy KMS statement and delegate to key policy — suggest doing the same here.

Thanks for feedback. 

1) Improved.

2)  The Resource is data.aws_kms_key.dashboard_cursor[0].arn, not the secret's kms_key_id — the aws_kms_key data source canonicalizes alias/key-ID references to the key ARN, so alias-created secrets apply cleanly.

haseebrabbani (approved):

PR #311, opened 2026-07-06
feat(server): release accounts on canonicalized guardian switch

### LINKED ISSUE #305: Guardian not notified of account deregistration on guardian_switch causes diverging state
**Description**
After a wallet switches to a new guardian, the old guardian continues to list and mutate state for that account. There’s no push mechanism to notify the old guardian it lost the account.

**Result**: Same accountId exists on both old and new guardians with different currentCommitment values. Both believe they’re authoritative.

**Repro Steps**

- Wallet A is served by Guardian X
- Wallet A submits guardian_switch tx to Guardian Y
- Check dashboards: Wallet A still appears on Guardian X and Guardian Y
- Both guardians accept state updates for Wallet A independently

**Impact**

- Split-brain: Two guardians hold conflicting state for one account
- Stale acceptance: Old guardian accepts txs after the account switched away. State seems to reset if wallet switches back, but behavior is undefined
- Unclear canonical source: With diverging commitments, it’s not obvious which one the network treats as valid

**Root Cause**

guardian_switch txs go to the new guardian only. Old guardian lacks any notification or required reconciliation against on-chain guardian metadata.

**Potential Approaches**

- Guardian polls chain for commitments it doesn’t recognize - works for small account sets, but polling cost scales poorly
- Node provides filtered events API - node sends only relevant deregistration events to guardians
- Push deregistration on switch - guardian_switch tx explicitly notifies old guardian
- Mark account disabled on detection - stop serving state once guardian realizes it’s stale

haseebrabbani:
## Summary

Fixes #305 — after a wallet switches guardians, the old guardian kept listing the account and accepting mutations, producing split-brain state.

Since #287, both multisig clients already push the `SwitchGuardian` delta to the **pre-switch** guardian, best-effort, so it canonicalizes there like any other delta. The missing piece was the server-side reaction — this PR adds it, per the approach discussed in the issue: **recognize-and-release, no new endpoint.**

## How it works

After a delta commits (optimistic mode) or canonicalizes (candidate mode — already verified against the on-chain commitment by the worker), the server compares the guardian public key commitment in the resulting state (same extraction `validate_guardian_commitment` uses in `/configure`) against its own ack key. On mismatch, the account transitions to a terminal **released** state:

- Mutating endpoints (`push_delta`, `push_delta_proposal`, `sign_delta_proposal`) refuse with `409 GUARDIAN_ACCOUNT_RELEASED` (gRPC `FailedPrecondition`), carrying `released_at`. Reads keep working so the wallet/operator can fetch final state.
- An `accounts.release` audit event records the new guardian commitment, delta nonce, and commitment.
- The dashboard account list/detail expose `released_at`.
- Only re-onboarding via `/configure` — which re-validates the guardian binding — reactivates the account. Released is orthogonal to pause, so an operator unpause can never resurrect a switched-away account; `/configure` clears released but preserves pause (deliberate, tested asymmetry).
- Like pause, released state is owned by dedicated `set_released`/`clear_released` metadata methods (filesystem + Postgres, with migration); generic metadata writes never change it.

Because the trigger is a commitment-bound, chain-verified delta, there is no new authenticated surface and nothing to replay — and the old guardian converges to the exact post-switch commitment instead of freezing stale, so dashboards on both guardians show the same commitment, one marked released.

## Testing

New e2e test (`--features e2e`) exercises the full production flow: a real `update_guardian_public_key` executed on a mock chain (mirroring `test_switch_guardian_server_reconstruction_matches_execution`), the client's exact `tx_summary.to_json()` payload pushed through the real `push_delta` service, the canonicalization worker verifying against the executed on-chain commitment, then the release lifecycle — released set, mutation refused, reads served, `/configure` reactivation. Unit tests cover the hook (mismatch/match/no-binding/error/EVM), the chokepoint (released wins over paused), first-writer-wins and set()-preservation semantics, and `/configure` clearing.

## Out of scope (follow-ups)

- TS operator-client typing for `released_at` / `account_released` (server changes are backward-compatible without it).
- Chain-reconciliation sweep for clients that never push (the push is best-effort by design — `ProposalResult::Offline` lets a switch proceed with an unreachable old guardian), and longer-term the node-side filtered-events API. The release transition takes any detection source, so both plug in later.

## Summary by CodeRabbit

* **New Features**
  * Added support for tracking when an account is released after switching guardians.
  * Dashboard account summaries and details now include a `released_at` timestamp.
  * Mutation responses now return a clear account-released error with timestamp details.
  * Re-onboarding via configure now restores released accounts.

* **Bug Fixes**
  * Released status is now preserved correctly during account updates.
  * Release state is handled consistently across dashboard, API, and storage paths.

* **Documentation**
  * Updated troubleshooting, concepts, and API schemas to reflect the new released-account behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3245,6 +3245,8 @@ dependencies = [
  "miden-protocol",
  "miden-rpc-client",
  "miden-standards",
+ "miden-testing",
+ "miden-tx",
  "pin-project-lite",
  "prost",
  "rand 0.9.2",
--- a/crates/server/Cargo.toml
+++ b/crates/server/Cargo.toml
@@ -72,6 +72,8 @@ diesel_migrations = { version = "2.2", optional = true }
 # Optional dependencies for e2e tests
 miden-client = { workspace = true, features = ["testing"], optional = true }
 miden-confidential-contracts = { path = "../contracts", optional = true }
+miden-testing = { workspace = true, optional = true }
+miden-tx = { workspace = true, optional = true }
 
 [build-dependencies]
 tonic-prost-build = { workspace = true }
@@ -102,4 +104,4 @@ postgres = [
     "tokio-postgres-rustls",
 ]
 integration = []
-e2e = ["miden-client", "miden-confidential-contracts"]
+e2e = ["miden-client", "miden-confidential-contracts", "miden-testing", "miden-tx"]
--- a/crates/server/migrations/2026-07-06-000001_account_released_field/down.sql
+++ b/crates/server/migrations/2026-07-06-000001_account_released_field/down.sql
@@ -1,0 +1,2 @@
+DROP INDEX IF EXISTS idx_account_metadata_released;
+ALTER TABLE account_metadata DROP COLUMN IF EXISTS released_at;
--- a/crates/server/migrations/2026-07-06-000001_account_released_field/up.sql
+++ b/crates/server/migrations/2026-07-06-000001_account_released_field/up.sql
@@ -1,0 +1,14 @@
+-- Issue #305: release-on-guardian-switch. `released_at` marks the UTC
+-- instant this server detected (via a canonicalized SwitchGuardian
+-- delta) that it is no longer the account's guardian. NULL while this
+-- server is the guardian. Terminal until the account re-onboards via
+-- /configure; owned by set_released / clear_released at the trait
+-- level — generic metadata writes never change it. Orthogonal to
+-- paused_at so an operator unpause cannot resurrect a released account.
+ALTER TABLE account_metadata ADD COLUMN released_at TIMESTAMPTZ NULL;
+
+-- Partial index supports "list all released accounts" cheaply; size
+-- scales with the count of released accounts, not total accounts.
+CREATE INDEX IF NOT EXISTS idx_account_metadata_released
+    ON account_metadata(released_at)
+    WHERE released_at IS NOT NULL;
--- a/crates/server/src/api/dashboard.rs
+++ b/crates/server/src/api/dashboard.rs
@@ -592,6 +592,7 @@ mod tests {
             last_auth_timestamp: None,
             paused_at: None,
             paused_reason: None,
+            released_at: None,
         };
         state
             .metadata
@@ -1130,6 +1131,7 @@ mod tests {
             last_auth_timestamp: None,
             paused_at: None,
             paused_reason: None,
+            released_at: None,
         }
     }
 
--- a/crates/server/src/api/dashboard_feeds.rs
+++ b/crates/server/src/api/dashboard_feeds.rs
@@ -453,6 +453,7 @@ mod tests {
             last_auth_timestamp: None,
             paused_at: None,
             paused_reason: None,
+            released_at: None,
         }
     }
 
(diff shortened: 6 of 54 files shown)


pepebndc (changes requested):
Requesting changes on two contract-completeness items before merge. The core release logic is correct and matches the recognize-and-release approach agreed in #305; both points below are about the repo's mandatory contract-propagation rules in AGENTS.md, not runtime behavior.

**[High] Missing client-parity updates (AGENTS.md Definition of Done #2 / §3.2)**

This PR changes two contract surfaces but touches no client packages:

- Dashboard surface: `released_at` was added to `DashboardAccountSummary` and `DashboardAccountDetail`. AGENTS.md §8.2 requires `/dashboard/*` shape changes to update `packages/guardian-operator-client` in the same PR.
- Per-account mutating endpoints (`push_delta`, `push_delta_proposal`, `sign_delta_proposal`) now return a new `409 GUARDIAN_ACCOUNT_RELEASED` carrying a `released_at` body; `packages/guardian-client` error-shape handling (`GuardianHttpError`) should cover it.

The "backward-compatible, defer to follow-up" rationale holds for wire compatibility, but AGENTS.md §3.2 treats server contract changes as multi-package by default. At minimum, please land the operator-client typing (`released_at` / `account_released`) in this PR.

**[Medium] `ApiErrorResponse` OpenAPI schema is missing `released_at`**

`ErrorResponse` now serializes `released_at` for `GUARDIAN_ACCOUNT_RELEASED`, but the hand-maintained `ApiErrorResponse` mirror in `crates/server/src/openapi.rs` (which documents `paused_at` / `paused_reason`) was not given a `released_at` field. Since that file is not in the diff, regeneration alone will not add it, and the committed `docs/openapi*.json` only got `released_at` on the dashboard account schemas, not on the error envelope. Please add the field to `ApiErrorResponse` and regenerate with `cargo run --features evm --bin gen-openapi -- docs`.

Everything else looks good: the guardian-key comparison is apples-to-apples with `validate_guardian_commitment`, reactivation is correctly gated behind commitment re-validation, storage backend parity is symmetric and first-writer-wins is atomic, released-wins-over-paused is correct, and the e2e drives the real push_delta plus canonicalization worker. A couple of low-severity notes (untested negative reactivation path, optimistic-mode trust) I will pass along separately.

haseebrabbani:
@pepebndc both items addressed in the latest commit:

**Client parity:**
- `guardian-operator-client`: `released_at` on the dashboard account models (wire + camelCase + strict parser), `account_released` in the `DashboardErrorCode` union with the `GUARDIAN_ACCOUNT_RELEASED` wire mapping in both envelope parsers (mirroring `account_paused`, including the contract-drift guards), and `AccountReleasedErrorDetails` exported.
- `guardian-client`: `GuardianHttpError` previously exposed only the raw body — it now parses the JSON envelope and exposes `code` and `releasedAt`, so callers can branch on `GUARDIAN_ACCOUNT_RELEASED` without string-matching. Tests cover the released envelope, other codes, and non-JSON bodies.
- `crates/client` (Rust): left unchanged deliberately — it surfaces `tonic::Status` generically including the `Status::details` JSON (which carries `code` + `released_at`), the same way `GUARDIAN_ACCOUNT_PAUSED` has always flowed through it.

**OpenAPI:** `ApiErrorResponse` in `openapi.rs` now carries `released_at` (and the `retryable` doc mentions the new code); all four specs regenerated and `gen-openapi --check` passes.

pepebndc (approved):
lgtm

PR #312, opened 2026-07-06
fix(server): release diverged canonicalization candidates instead of …

### LINKED ISSUE #303: consume_notes canonicalization stalls minutes under concurrent same-account proposals (verify_state commitment mismatch burns the 600s grace) + 409 ConflictPendingDelta churn
## Summary

When a guardian-co-signed account performs **concurrent / rapidly-overlapping** transactions (e.g. a `consume_notes` while the same account is also being mutated by another in-flight tx), canonicalization of the pending delta **stalls for minutes** — up to the full `submission_grace_period_seconds` (600s in prod) — and the account is locked the whole time, returning `409 ConflictPendingDelta` to every new proposal. Serializing operations per account makes the problem disappear entirely.

This was found while integrating the guardian into the Miden Wallet (hosted `guardian.openzeppelin.com`, Miden 0.15 **testnet**, `@openzeppelin/guardian-client@0.15.0-rc.0`). Code references below are against `guardian@89591a3` (current `main`).

## Measurements (these rule out the obvious suspects)

- Testnet block production: **~2.9 s/block** (fast).
- Guardian HTTP latency: `/pubkey` ~0.2 s; every `/state`, `/delta/proposal`, `/delta` request **~125–170 ms** (fast).
- Yet **consume finalization** (time from `POST /delta/proposal` until the co-signed delta is canonical / visible via `/state`) under a concurrent workload: p50 ~5.5 s but **p90 = 216 s, max = 586 s**, and **13% of `POST /delta/proposal` returned `409`**.

So it's neither network latency nor block time — it's canonicalization waiting on a commitment that never matches.

## Root cause (as I read the source)

1. `POST /delta` makes the delta a **Candidate** and sets `has_pending_candidate = true`. Any subsequent `POST /delta/proposal` for that account is rejected with **`409 ConflictPendingDelta`** — `crates/server/src/services/push_delta_proposal.rs:79` (`has_pending_candidate`) → `:91` (`return Err(GuardianError::ConflictPendingDelta)`).
2. The canonicalization worker (every `check_interval_seconds = 10`) calls `verify_state` → `get_account_commitment()` RPC and finalizes **only when the on-chain commitment equals the guardian's locally-computed expected commitment** — `crates/server/src/network/miden/mod.rs:108`, `crates/server/src/jobs/canonicalization/processor.rs:151-171`.
3. On a mismatch **within the grace period**, it just defers without consuming retry budget — `processor.rs:174-189` (`candidate_age_seconds < self.submission_grace_period_seconds` → *"Delta verification failed during submission grace period; will retry without consuming retry budget"*). Prod config: `CanonicalizationConfig::new(10, 48).with_submission_grace_period_seconds(600)` — `crates/server/src/main.rs:40`.

The trap: the expected commitment is computed for **base_state + this_delta**. If the account legitimately advances on-chain via *another* tx before this candidate is verified, the actual commitment is `base_state + other_delta(s) + this_delta` (or a different ordering) and **never equals the guardian's expectation** — so the candidate burns the entire 600 s grace and is then discarded, blocking the account (409) the whole time.

## Controlled experiment (concurrency is the trigger)

Identical network + hosted guardian; only the wallet's operation pattern changed:

| metric | concurrent workload | serialized (one op per account at a time) |
|---|---|---|
| `409 ConflictPendingDelta` | 16 / 127 (13%) | **0 / 32 (0%)** |
| `/delta/proposal` finalize gap p90 / max | 216 s / 586 s | **43.5 s / 57.4 s** |
| `/state` polls (client waiting) | 1,879 | 298 |
| result | crawled, never converged | **passed, balances conserved, settled in 4 s** |

## Suggested direction

When `verify_state` fails because the account has **advanced past** the expected commitment (vs. a genuine verification failure), canonicalization could re-derive the expected commitment from the *current* on-chain account state and re-check, rather than waiting out the grace period against a now-stale expectation. At minimum, distinguishing "account moved on / superseded" from "invalid" would let the server release `has_pending_candidate` quickly instead of holding the account locked for up to 600 s. Happy to provide the full request/timeline traces.

## Environment
- guardian: hosted `guardian.openzeppelin.com`; source `main@89591a3`
- client: `@openzeppelin/guardian-client@0.15.0-rc.0`, `@openzeppelin/miden-multisig-client`
- chain: Miden 0.15 testnet (`rpc.testnet.miden.io`, blocks ~2.9 s)

haseebrabbani:
Fixes #303

## Problem

When a guardian-cosigned account advances on-chain past the state a candidate delta was built on, the worker's expected commitment (`stored_base + this_delta`) can never match on-chain again. The candidate burned the full 600s submission grace, then 48 retries (~480s more), while `has_pending_candidate` kept the account locked — every new proposal got `409 conflict_pending_delta` for up to ~18 minutes

## Root cause

`verify_state` returned an opaque `Err` for *any* mismatch, so the worker could not distinguish "transaction hasn't landed yet" (grace is the right response) from "the account moved past this candidate's base" (waiting can never succeed). Both cases deferred until grace + retry budget were exhausted.

## Fix

`NetworkClient::verify_state` now returns a typed outcome — `Match` / `Mismatch { on_chain }` — with `Err` reserved for failures to make the comparison at all (RPC error, malformed state). The canonicalization worker classifies a mismatch against the candidate's own commitments:

| on-chain commitment | meaning | action |
|---|---|---|
| `== prev_commitment` | tx not landed yet | defer within grace, then consume retry budget (unchanged) |
| `== expected new commitment` | tx landed | canonicalize (unchanged) |
| neither | account advanced past the candidate's base; its tx is anchored to `prev_commitment` and can never land | discard, delete matching proposal, clear `has_pending_candidate` — account released immediately |

The divergence discard bypasses the grace period, collapsing the lock from ~18 minutes to ~2 worker ticks (~20 seconds).

## Safety measures

- **Stale-read shield**: discarding requires `divergence_confirmations` (default 2) *consecutive* diverged observations. The counter is persisted as `divergence_count` on `DeltaStatus::Candidate` (`#[serde(default)]`, so existing JSONB rows deserialize unchanged), so confirmation survives worker restarts. A single read from a lagging RPC node cannot discard a candidate whose tx actually landed.
- **Full release, not a flag flip**: the diverged path reuses the same discard routine as the max-retries path (delete delta + delete matching proposal + clear flag), so no zombie candidate rows and no stranded pending proposals that would let clients re-submit a stale intent.
- **RPC errors don't classify**: only a *successful* read that matches neither commitment counts toward divergence; fetch failures keep today's grace/retry behavior.
- `TestDeltasProcessor` uses `u32::MAX` confirmations, preserving its never-discard semantics for integration/e2e tests.

## Observability

New outcome labels on `guardian_canonicalization_candidates_total`: `diverged` (released) and `divergence_deferred` (observed once, awaiting confirmation). `spec/processes.md` updated to describe the three-way classification.

## Testing

- Four new worker tests: on-chain-at-base defers within grace with no writes; first diverged observation persists the counter without discarding; second observation discards and releases inside the grace window; `divergence_confirmations = 1` discards immediately.

## Residual limitations (out of scope)

- A candidate whose tx **never lands at all** (e.g. the wallet's tx failed locally after acking — see 0xMiden/wallet#313) leaves on-chain at `prev_commitment`, indistinguishable from slow proving, so it still holds the account for grace + retries. Releasing that faster would need a client-initiated abandon/withdraw endpoint — happy to file a follow-up.
- A *permanently* lagging RPC node could produce two consecutive stale reads and discard a landed candidate, leaving the guardian's stored state behind on-chain until the client re-pushes the landed delta. Such a node breaks canonicalization today as well.

## Summary by CodeRabbit

* **New Features**
  * Added configurable divergence confirmation handling during canonicalization, letting the system wait for repeated divergence before discarding a candidate.
  * Improved state-check reporting so mismatches are tracked separately from successful matches.

* **Bug Fixes**
  * Candidate items now preserve and expose divergence progress across retries.
  * Divergence-related outcomes are now reflected more accurately in metrics and dashboard-facing data.

* **Tests**
  * Updated and expanded test coverage for retry, grace-period, and divergence scenarios.

### DIFF
--- a/crates/server/src/builder/canonicalization.rs
+++ b/crates/server/src/builder/canonicalization.rs
@@ -13,6 +13,14 @@ pub struct CanonicalizationConfig {
 
     /// Minimum age a candidate must reach before verification failures consume retry budget.
     pub submission_grace_period_seconds: u64,
+
+    /// Consecutive worker ticks that must observe the on-chain commitment at
+    /// neither the candidate's previous nor its expected new commitment before
+    /// the candidate is discarded as diverged (the account advanced past the
+    /// state it was built on, so it can never verify). Values above 1 shield
+    /// against acting on a single stale RPC read; the divergence discard
+    /// bypasses the submission grace period.
+    pub divergence_confirmations: u32,
 }
 
 impl Default for CanonicalizationConfig {
@@ -21,6 +29,7 @@ impl Default for CanonicalizationConfig {
             check_interval_seconds: 10,           // Try every 10 seconds
             max_retries: 18,                      // 18 attempts (total: ~3 minutes)
             submission_grace_period_seconds: 600, // Allow proving/submission to settle first
+            divergence_confirmations: 2,          // Two ticks to rule out a stale read
         }
     }
 }
@@ -32,6 +41,7 @@ impl CanonicalizationConfig {
             check_interval_seconds,
             max_retries,
             submission_grace_period_seconds: Self::default().submission_grace_period_seconds,
+            divergence_confirmations: Self::default().divergence_confirmations,
         }
     }
 
@@ -49,5 +59,12 @@ impl CanonicalizationConfig {
     /// Get submission grace period as Duration
     pub fn submission_grace_period(&self) -> Duration {
         Duration::from_secs(self.submission_grace_period_seconds)
+    }
+
+    /// Override the number of consecutive diverged observations required
+    /// before a candidate is discarded.
+    pub fn with_divergence_confirmations(mut self, confirmations: u32) -> Self {
+        self.divergence_confirmations = confirmations;
+        self
     }
 }
--- a/crates/server/src/builder/startup.rs
+++ b/crates/server/src/builder/startup.rs
@@ -149,6 +149,7 @@ mod tests {
                 check_interval_seconds: 10,
                 max_retries: 48,
                 submission_grace_period_seconds: 600,
+                divergence_confirmations: 2,
             }),
             3,
             true,
--- a/crates/server/src/delta_object.rs
+++ b/crates/server/src/delta_object.rs
@@ -25,9 +25,10 @@ pub enum DeltaStatus {
         /// Consecutive canonicalization ticks that observed the on-chain
         /// commitment at neither this delta's `prev_commitment` nor its
         /// expected new commitment — i.e. the account advanced past the
-        /// state this candidate was built on. Requiring more than one
-        /// observation before discarding shields against a single stale
-        /// RPC read.
+        /// state this candidate was built on. Reset to zero whenever a
+        /// read shows the account still at the candidate's base, so only
+        /// an unbroken streak counts. Requiring more than one observation
+        /// before discarding shields against stale RPC reads.
         #[serde(default)]
         divergence_count: u32,
     },
@@ -141,6 +142,21 @@ impl DeltaStatus {
                 timestamp: timestamp.clone(),
                 retry_count: *retry_count,
                 divergence_count: divergence_count + 1,
+            },
+            _ => self.clone(),
+        }
+    }
+
+    pub fn with_reset_divergence(&self) -> Self {
+        match self {
+            Self::Candidate {
+                timestamp,
+                retry_count,
+                divergence_count: _,
+            } => Self::Candidate {
+                timestamp: timestamp.clone(),
+                retry_count: *retry_count,
+                divergence_count: 0,
             },
             _ => self.clone(),
         }
--- a/crates/server/src/delta_object.rs
+++ b/crates/server/src/delta_object.rs
@@ -22,6 +22,14 @@ pub enum DeltaStatus {
         timestamp: String,
         #[serde(default)]
         retry_count: u32,
+        /// Consecutive canonicalization ticks that observed the on-chain
+        /// commitment at neither this delta's `prev_commitment` nor its
+        /// expected new commitment — i.e. the account advanced past the
+        /// state this candidate was built on. Requiring more than one
+        /// observation before discarding shields against a single stale
+        /// RPC read.
+        #[serde(default)]
+        divergence_count: u32,
     },
     Canonical {
         timestamp: String,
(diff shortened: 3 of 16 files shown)


pepebndc (changes requested):
Requesting one change before merge. The typed three-way classification (`Match` / `Mismatch { on_chain }`, `Err` reserved for genuine comparison failure) is the right fix for #303, and reusing the full `remove_candidate` routine on the diverged path is clean. This is about the confirmation-counter contract.

**[Medium] `divergence_count` is never reset, so "consecutive" is really "cumulative"**

The field doc and the PR description both say *consecutive* diverged observations, but nothing resets the counter on a non-diverged read: `with_incremented_retry` preserves `divergence_count`, and the grace-defer path writes nothing. A `diverged(stale) -> prev/grace -> diverged(stale)` interleaving therefore accumulates toward the threshold and can discard a candidate whose tx is still in flight, which is the exact stale-read case the confirmation gate is meant to prevent. Low probability (a truly current read returns `Match` and ends the loop), but it contradicts the stated contract and is untested.

Fix: reset the persisted `divergence_count` to 0 on any non-diverged observation (note the grace-defer branch currently does no write, so strict "consecutive" costs one extra write there), or relabel the contract as "cumulative". Either way, please add a `diverged -> base -> diverged` test.

The rest is solid: serde default keeps existing candidate rows deserializing unchanged, RPC errors correctly do not classify, `TestDeltasProcessor` preserves never-discard semantics, and metrics plus spec/processes.md are updated. A few low/nit items (permanent lock if the flag-clear write fails after the delta is deleted, `divergence_confirmations = 0` guard, dashboard config surfacing) I will pass along separately.

haseebrabbani:
@pepebndc Good catch — now fixed:

- A successful read that shows on-chain still at the candidate's base now resets the persisted streak (`reset_divergence_streak`, no-op/no write unless a streak was in progress, so the common path costs no extra write).
- RPC errors intentionally do *not* reset: no observation was made, so the streak is left untouched — a flaky node can neither advance nor erase divergence evidence.
- Added the `diverged -> base -> diverged` test: persisted counts go `[1, 0, 1]` across three ticks, no discard, lock intact.

pepebndc (approved):
lgtm

PR #314, opened 2026-07-08
fix(multisig-client): recover/verify ECDSA public key in MidenWalletSigner instead of using the commitment

### LINKED ISSUE #313: MidenWalletSigner uses the account commitment as the ECDSA public key → tx fails at execute with "invalid public key commitment"
### Summary

For an ECDSA (secp256k1 / keccak) guardian cosigner backed by a wallet, `MidenWalletSigner` (`@openzeppelin/miden-multisig-client`) can expose the 32-byte **account commitment** as its `publicKey` instead of the real 33-byte compressed ECDSA key. That value is then packed into the transaction's advice map, and the Miden VM kernel rejects it at **execute** time:

```
assertion failed with error message: invalid public key commitment
```

Signing itself appears to succeed — the failure only surfaces when the transaction executes.

### Root cause

`MidenWalletSigner`'s constructor set:

```ts
this.publicKey = publicKey ?? localAuthSigner?.publicKey ?? commitment;
```

When an ECDSA wallet cosigner is constructed **without** an explicit `publicKey` and **without** a `localAuthSigner`, `publicKey` silently falls back to `commitment` (32 bytes). The transaction kernel (`crypto/dsa/ecdsa_k256_keccak.masm`) loads the advertised public key, Poseidon2-hashes it, and asserts it equals the account's stored `pub_key_commitment`:

```
assert_eqw.err="invalid public key commitment"
```

Since `Poseidon2(commitment) != commitment`, the assertion fails. What's required is a 33-byte key whose Poseidon2 hash equals the commitment; the 32-byte commitment itself can never satisfy that.

`ParaSigner` validates its key via `EcdsaFormat.validatePublicKeyHex`, so the Para flow works with a proper 33-byte key. `MidenWalletSigner` had no equivalent guard.

### Where it surfaces

`buildGuardianSignatureFromSigner` (`multisig/signing.ts`) reads `signer.publicKey` to build the `ProposalSignature` sent to the guardian and packed into the advice map, so the bad value flows straight from the signer into the transaction.

### Impact

Wallet-delegated ECDSA guardian cosigners that don't pass an explicit public key cannot transact — every transaction traps at execute. Falcon is unaffected (there `publicKey === commitment` is legitimate).

### Expected behaviour

`MidenWalletSigner` must never advertise the commitment as an ECDSA public key. It should resolve the real key — from an explicit argument (validated), a delegate signer, or by recovering it from the signature — and verify that key against the commitment before using it.

### Reproduction

A real-crypto repro (real secp256k1 keypair + real WASM Poseidon2, no mocks) is attached in the fixing PR. It builds a wallet-delegated ECDSA `MidenWalletSigner` with no explicit key and asserts the resolved `publicKey` Poseidon2-hashes to the account commitment:

- **current code:** `signer.publicKey` is the 32-byte commitment `0x09b3…` → `Poseidon2(publicKey) != commitment` → kernel `assert_eqw` traps.
- **expected:** `signer.publicKey` is the real 33-byte key `0x034f…` → `Poseidon2(publicKey) == commitment`.

### Environment

- Package: `@openzeppelin/miden-multisig-client`
- Signer: ECDSA (`EcdsaK256Keccak`) guardian cosigner, wallet-delegated (`WalletSigningContext.signBytes`)

WiktorStarczewski:
Fixes #313.

## Problem

For an ECDSA guardian cosigner backed by a wallet, `MidenWalletSigner` exposed the 32-byte account **commitment** as its `publicKey` instead of the real 33-byte ECDSA key. The constructor did:

```ts
this.publicKey = publicKey ?? localAuthSigner?.publicKey ?? commitment;
```

With no explicit key and no `localAuthSigner`, an ECDSA cosigner fell back to `commitment`. That value is packed into the transaction advice map, and the kernel (`crypto/dsa/ecdsa_k256_keccak.masm`) asserts `Poseidon2(publicKey) == pub_key_commitment` → `assert_eqw "invalid public key commitment"`. Since `Poseidon2(commitment) != commitment`, the tx traps at **execute**. Signing succeeds, so the failure only shows at execute time. `ParaSigner` guards this with `EcdsaFormat.validatePublicKeyHex`; `MidenWalletSigner` did not.

## Fix

For ECDSA, `publicKey` is now resolved (never the commitment) as:

1. an explicit key argument — validated (33/65 bytes) and compressed, like `ParaSigner`;
2. a delegate `localAuthSigner`'s key; or
3. for a pure wallet signer, **recovered from the signer's own signature** (secp256k1 recovery via the already-present `@noble/curves`) and **verified against the commitment** — the same `Poseidon2(publicKey) == commitment` invariant the kernel enforces — failing closed on mismatch.

Falcon is unchanged (`publicKey === commitment` is legitimate there). `publicKey` is now a lazily-resolved getter; every reader reads it after signing (verified across both `miden-multisig-client` and `guardian-client`), so no consumer regresses. New helper: `EcdsaFormat.recoverCompressedPublicKeyHex`.

## Repro — fails without the fix, passes with it

The PR is structured so the repro is a real, runnable regression, not just prose:

- **`6395b91` (test only, no fix)** — the repro is RED.
- **`cc780a5` (fix)** — GREEN.
- **`ff71139`** — adds an integration-level regression at the guardian seam (`buildGuardianSignatureFromSigner`).

`miden-wallet.ecdsa-recovery.test.ts` uses a real secp256k1 keypair and the real WASM Poseidon2 (no mocks), reproducing the exact kernel invariant:

| | before fix | after fix |
|---|---|---|
| `signer.publicKey` | `0x09b3…` (32-byte commitment) | `0x034f…` (real 33-byte key) |
| `Poseidon2(publicKey) == commitment` | ❌ kernel `assert_eqw` traps | ✅ |

Coverage also includes: fail-closed on a mismatched key, explicit-key validation, 65-byte → compressed, recovery via a non-`signCommitment` path, and the guardian `ProposalSignature` integration test.

## Tests

- `packages/miden-multisig-client`: 6/6 recovery tests; 319 pass across the package; `tsc --noEmit` clean. (2 unrelated failures in `procedure-roots.test.ts` are environmental — they shell out to `cargo`/`protoc`.)
- Note: this repo's CI (`ci.yml`) is Rust-only and does not run the JS package vitest suites, so the red→green is demonstrated by the commit structure above rather than enforced by CI. Adding a JS vitest job is a reasonable separate follow-up.

## Scope

Self-contained in `@openzeppelin/miden-multisig-client`; `guardian-client` untouched (the `Signer` contract is unchanged — `publicKey` stays `readonly string`, just lazily populated). No accompanying PR required; a `miden-wallet` dependency bump is the only follow-up after publish.

## Summary by CodeRabbit

* **New Features**
  * Improved ECDSA signer behavior to recover and reuse the wallet’s public key automatically when it isn’t provided upfront.
  * Added support for recovering a compressed public key from standard ECDSA signatures, including Ethereum-style recovery bytes.

* **Bug Fixes**
  * Strengthened validation for invalid or mismatched ECDSA keys and signatures.
  * Ensured signer output now uses the recovered public key when building guardian signatures.

### DIFF
--- a/packages/miden-multisig-client/src/signers/miden-wallet.ecdsa-recovery.test.ts
+++ b/packages/miden-multisig-client/src/signers/miden-wallet.ecdsa-recovery.test.ts
@@ -3,6 +3,7 @@ import { secp256k1 } from '@noble/curves/secp256k1';
 import { keccak_256 } from '@noble/hashes/sha3.js';
 
 import { MidenWalletSigner, type WalletSigningContext } from './miden-wallet.js';
+import type { Signer } from '../types.js';
 import { tryComputeEcdsaCommitmentHex } from '../utils/signature.js';
 import { bytesToHex, hexToBytes } from '../utils/encoding.js';
 import { buildGuardianSignatureFromSigner } from '../multisig/signing.js';
@@ -10,16 +11,26 @@ import { buildGuardianSignatureFromSigner } from '../multisig/signing.js';
 // Real-crypto repro for the reported failure:
 //   "assertion failed with error message: invalid public key commitment"
 //
-// This test uses the real WASM SDK (Poseidon2/PublicKey, via tests/setup-wasm.ts)
-// and a real secp256k1 keypair — NOT mocks — so it reproduces the exact kernel
+// These tests use the real WASM SDK (Poseidon2/PublicKey, via tests/setup-wasm.ts)
+// and real secp256k1 keypairs — NOT mocks — so they exercise the exact kernel
 // invariant `Poseidon2(publicKey) == pub_key_commitment` that the transaction
 // kernel enforces in `ecdsa_k256_keccak.masm` (`assert_eqw "invalid public key
 // commitment"`). A wallet-delegated ECDSA cosigner whose `publicKey` degrades to
-// the 32-byte commitment fails that invariant.
+// the 32-byte commitment — or is a valid-but-wrong key — fails that invariant.
 
-// Deterministic key so the repro is reproducible run-to-run.
+// Deterministic keys so the repro is reproducible run-to-run.
 const PRIVATE_KEY = hexToBytes('0x' + '11'.repeat(32));
 const COMPRESSED_PUBKEY_HEX = bytesToHex(secp256k1.getPublicKey(PRIVATE_KEY, true));
+
+// A second, unrelated key: valid secp256k1 but NOT the one the commitment is for.
+const OTHER_PRIVATE_KEY = hexToBytes('0x' + '22'.repeat(32));
+const OTHER_COMPRESSED_PUBKEY_HEX = bytesToHex(secp256k1.getPublicKey(OTHER_PRIVATE_KEY, true));
+
+/** Asserts a nullable value is present and returns it narrowed (avoids `!`). */
+function expectPresent<T>(value: T | null | undefined): T {
+  expect(value ?? null).not.toBeNull();
+  return value as T;
+}
 
 /**
  * A wallet that signs like the real Miden Wallet ECDSA path: it keccak256-hashes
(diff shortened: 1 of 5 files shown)


WiktorStarczewski:
## Addressed the review findings (pushed `971d5d6`)

Verified each finding against the code before acting.

**🟠 `slice(1)` drops the first byte / unprefixed 65-byte signatures (`miden-wallet.ts`)** — resolved as a doc fix, **not** the length heuristic. This package's wallet-signature contract is *always* 1-byte-scheme-tag-prefixed: `signatureHexToBytes` prepends the tag, the SDK's own `EcdsaSigner`/`FalconSigner` do the identical unconditional `serialize().slice(1)`, and the post-slice bytes are what get sent to the guardian — so an unprefixed input would corrupt *every* ECDSA signature, not just recovery. The proposed `length === 65 ? keep : slice(1)` heuristic would convert the clean fail-loud for a prefixed 64-byte (no-recovery) signature into a silent misread, so I kept the unconditional `slice(1)` and instead tightened the misleading `WalletSigningContext` JSDoc (dropped "optionally" — the tag is mandatory).

**🎯 Validate the resolved ECDSA key before caching (`miden-wallet.ts`)** — applied, and hardened further. Explicit, delegate, and recovered keys now all pass through one `Poseidon2(key) == commitment` gate. A follow-up self-review caught that `tryComputeEcdsaCommitmentHex` also returns `null` for a valid-*length* but off-curve key (not only when the SDK is cold), which the best-effort skip would have let through — so explicit/delegate keys are now also validated as real secp256k1 points (SDK-independent, via `@noble/curves`), and the cold-SDK skip no longer caches (verification stays idempotent). New helper `EcdsaFormat.isValidPublicKeyPoint`.

**🎯 Explicit-key test used a length-shaped dummy + fake commitment (`miden-wallet.test.ts`)** — applied. The two SDK-mocked constructor tests that asserted `.publicKey` moved to the real-crypto suite (`miden-wallet.ecdsa-recovery.test.ts`) with real key/commitment fixtures; added match, mismatch, and off-curve/point-validity coverage for both explicit and delegate keys.

**🧹 Inline comments / non-null assertions** — applied. Removed the inline `//` comments from the implementation (AGENTS.md §12) and dropped the `!` assertions from the recovery tests via an `expectPresent` helper.

Net: 325 package tests pass, `tsc --noEmit` clean. (The 2 `procedure-roots` failures are environmental — that suite shells out to `protoc`.)

WiktorStarczewski:
I confirm that I have read and hereby agree to the OpenZeppelin Contributor License Agreement

zeljkoX (approved):
Thanks for identifying and creating fix!

LGTM!

Issue #320, opened 2026-07-15
Per-account proposal handshake cost grows with interleaved proposal traffic (send + consume) — +128–153% over 300 ops


Ivanlomoljo26:
### Environment
- `guardian.openzeppelin.com` v0.15.1 (`git 82225104c8a2`), Miden testnet
- Paired A/B stress runs: identical PRNG seed, 300 guardian-cosigned sends, per-transaction stage timing. Single variable: the receiving wallet either also issued consume (claim) proposals for ~50% of received notes, or none at all.

### Observation
With interleaved consumes, the guardian handshake stages grow monotonically over the run; removing consume traffic flattens exactly those stages:

| stage | growth Q1→Q5, claims on | claims off |
|---|---|---|
| proposal-create → signing | **+153%** | +77% (non-monotonic) |
| signing → sending | **+128%** | +41% |
| sending → submitting | **+43%** | **0%** |

Everything else was measured flat in both runs (client sync, proving, block wait, local execution, UI). Net effect: median send latency grows from ~11 s to ~19 s across the run, driven by these stages. Since the flattening tracks the traffic-mix knob and not client state, session age (survives page reloads), or network (per-call durations flat), the growth follows **per-account proposal co-traffic**.

This looks like the non-catastrophic sibling of #303: #312 fixed the multi-minute stall/409-churn form; the steady per-account cost growth under normal interleaved traffic remains.

### Where the time goes (narrowing measurements)
- **Server per-call response times are flat.** Guardian `delta`/`proposal` endpoints answered in ~250–280 ms at the start AND end of the runs; sync RPCs pinned at 201–215 ms. The growth is not server response latency per request.
- **The number of round-trips per proposal grows.** In a heavier run on the same account shape (2026-07-13, 442 ops), client-issued guardian proposal-path calls grew from 7.5 to 12.2 per operation while per-call durations stayed flat — the handshake takes more exchanges, not slower ones.
- **Account payload grows.** `GetAccount` was the only RPC whose duration grew (305 → 483 ms) over that run, tracking a growing account state payload.
- **It is not client-session state.** Full page reloads of both wallets at ops 100 and 200 produced no reset (+0.8% / +5.9% across the boundaries) — whatever accumulates is durable/server-side, not in-memory client state.
- **It is not queue depth at send time.** The sender's local tx queue was empty at the start of nearly every op; latency ramps even for ops issued with zero in-flight work.

### Suspected mechanism (falsifiable)
[...]

PR #326, opened 2026-07-17
perf: canonicalization performance

zeljkoX:
This PR adds bounded parallel account processing, candidate-only storage reads, improved RPC concurrency, canonicalization metrics, and benchmark sampling/reporting.

Canonicalization now recovers anomalous or legacy candidates whose stored new_commitment is missing or incorrect: after the server-recomputed state is verified on-chain, the verified commitment is persisted and the discrepancy is logged and counted. Normal push ingestion already stores the server-recomputed commitment.

Benchmark: `apply_delta` scales linearly with serialized account size because reconstruction deserializes and reserializes the unbounded executed_transactions replay map. Investigate incremental account reconstruction and a bounded or compressed replay-protection representation while preserving replay safety and existing-account compatibility for the upcoming PRs.

## Summary by CodeRabbit

* **New Features**
  * Canonicalization now supports bounded parallel processing, configurable through `GUARDIAN_CANONICALIZATION_MAX_CONCURRENT_ACCOUNTS`.
  * Added detailed canonicalization outcomes, including partial, cancelled, stale-base, and verification-failure states.
  * Added metrics for processing activity, commitment mismatches, candidate age, and pass outcomes.
  * Production benchmarks now optionally sample canonicalization wait times and include percentile reports and artifacts.
* **Bug Fixes**
  * Improved promotion safety during concurrent updates and guardian leadership changes.
  * Added RPC request timeouts and more reliable candidate-delta retrieval.
* **Documentation**
  * Updated configuration, troubleshooting, benchmark, API, and operational guidance.


zeljkoX:
Improves #320

haseebrabbani:
Two comments:

1. **Verified-commitment promotion is a semantics change worth calling out.** A candidate with a missing `new_commitment` previously never canonicalized; now promotion proceeds with the server-recomputed value once verification proves it on-chain, and a missing/differing client claim is only logged + counted. I think that's the right call, but it changes `new_commitment` on canonical records from client-attested to server-verified — worth confirming no downstream consumer relies on the former, and a line in the PR description.

2. **`apply_delta` is blocking CPU work now running at concurrency 50.** Under the old `Mutex` it was implicitly serialized; with `buffer_unordered` at the prod setting, up to 50 synchronous state-reconstruction calls can occupy tokio worker threads at once and stall API traffic on the leader replica. Suggest wrapping it in `spawn_blocking`, or measuring a worst-case `apply_delta` and documenting why it's negligible.

zeljkoX:
> Two comments:
> 
> 1. **Verified-commitment promotion is a semantics change worth calling out.** A candidate with a missing `new_commitment` previously never canonicalized; now promotion proceeds with the server-recomputed value once verification proves it on-chain, and a missing/differing client claim is only logged + counted. I think that's the right call, but it changes `new_commitment` on canonical records from client-attested to server-verified — worth confirming no downstream consumer relies on the former, and a line in the PR description.
> 2. **`apply_delta` is blocking CPU work now running at concurrency 50.** Under the old `Mutex` it was implicitly serialized; with `buffer_unordered` at the prod setting, up to 50 synchronous state-reconstruction calls can occupy tokio worker threads at once and stall API traffic on the leader replica. Suggest wrapping it in `spawn_blocking`, or measuring a worst-case `apply_delta` and documenting why it's negligible.

Thanks for the feedback.

Confirmed that downstream consumers treat new_commitment as the resulting canonical state commitment, not as a client attestation. Promotion now persists the server-recomputed commitment only after it has been verified on-chain. A missing or differing client claim is logged and counted. I’ve called out this semantics change in the PR description.

Benchmarks confirmed that apply_delta is material CPU work and grows with account size. I moved reconstruction off Tokio worker threads using spawn_blocking and placed both canonicalization and push_delta behind a shared process-wide concurrency bound. I also opened [#328](https://github.com/OpenZeppelin/guardian/issues/328) to track eliminating the underlying repeated full-account deserialization and sparse-Merkle-tree reconstruction.

haseebrabbani (approved):

haseebrabbani:
One thing before merge: capacity 1 serializes every `push_delta` reconstruction process-wide, where previously concurrent pushes reconstructed independently. With your bench showing up to ~1s per large account, one such job head-of-line-blocks all pushes, and sustained throughput caps at ~1/mean-reconstruction-cost — potentially tight against the 500 TPS target. Could we get one burst-scale benchmark run on this commit to confirm `push_delta/s` doesn't regress vs the pre-gate commit? If it does, a capacity of ~`num_cpus/2` for the CPU gate (keeping background admission at 1) would preserve the latency protection without the throughput ceiling.

zeljkoX:
> One thing before merge: capacity 1 serializes every `push_delta` reconstruction process-wide, where previously concurrent pushes reconstructed independently. With your bench showing up to ~1s per large account, one such job head-of-line-blocks all pushes, and sustained throughput caps at ~1/mean-reconstruction-cost — potentially tight against the 500 TPS target. Could we get one burst-scale benchmark run on this commit to confirm `push_delta/s` doesn't regress vs the pre-gate commit? If it does, a capacity of ~`num_cpus/2` for the CPU gate (keeping background admission at 1) would preserve the latency protection without the throughput ceiling.

Ran the A/B: burst-scale shape (ECDSA push-only, retire-after-first-push) against local release builds of this commit vs 6b34870, fresh DB per leg, canonicalization worker live, 1024 (×3) and 4096 (×2) concurrent users.

No regression — means within noise (1024-user: 1185 vs 1170 push/s; 4096-user: 1258 vs 1103 with overlapping ranges, ±15% run variance), zero failed pushes, all runs >1000 push/s. Burst-scale accounts are fresh by construction, so reconstruction is sub-ms and the serialized ceiling never binds at this shape.

Your head-of-line point does stand for large long-lived accounts, which the harness can't represent — and since the fix is free, I took the suggestion anyway: CPU gate is now max(1, available_parallelism()/2), background admission stays 1. Identical on 2-vCPU prod (resolves to 1); pays off on bigger hosts. A third bench leg on the dynamic build measured the same (1144/1129 push/s means).

PR #327, opened 2026-07-17
abandon candidate endpoint

### LINKED ISSUE #319: Stranded candidate (tx never lands) locks the account for grace+retries — follow-up to #312: client-initiated abandon/release
### Problem
#312 (shipped in v0.15.1) releases *diverged* canonicalization candidates quickly, but its "residual limitations" section documents the remaining case: a candidate whose transaction **never lands at all** still holds the account for the full submission grace + retry budget (~600 s + 48 retries ≈ 18 min). During that window every new proposal on the account is answered `409 conflict_pending_delta`.

We hit this case reproducibly on v0.15.1 (i.e. post-#312):

- In a 300-send run (2026-07-14), 845 total 409s arrived in on/off **single-account episodes of 4–16 min** — under the ~18 min grace+retry ceiling.
- The largest episode (12 consecutive operations) began **immediately after two transactions died client-side after guardian approval** (RPC submit failures) — approved candidates whose tx never landed.
- The wallet's local tx queue was empty at the start of 48/56 conflict windows — the pending delta is guardian-side state, not a queued client tx.
- Client impact: our wallet currently retries a 409 twelve times (~65 s) and then fails the tx; 56 sends failed this way in one run, and ops that caught 3–9 conflicts survived but formed the p99 latency tail (37–81 s).

One major *source* of never-landing candidates was just fixed in #314 (ECDSA signer bug → tx traps at execute after approval), but any post-approval client-side death (prover timeout, RPC submit error, crash) recreates the situation.

### Proposed fix
The follow-up #312's author already suggested: a **client-initiated abandon/release endpoint**, so a wallet that knows its tx died can free the candidate immediately instead of the account staying locked for grace+retries.

### References
#312 (residual-limitations section) · #303 · #313 / #314 · #316 (the inverse case — force canonicalization when the client knows the tx landed)

haseebrabbani:
Closes #319

## Problem

A candidate whose transaction dies client-side after approval (RPC submit failure, prover timeout, crash) looks identical to one that is slowly proving — on-chain stays at `prev_commitment`, so neither the grace period nor the #312 divergence release helps. The account stays locked for grace + retries (~18 min), with every new proposal answered `409 conflict_pending_delta`. Only the client knows the transaction is dead.

## What's new

**`POST /delta/candidate/abandon`** / gRPC **`AbandonDeltaCandidate`** — signature-authenticated (same header model as `push_delta_proposal`), body `{account_id, nonce}`; the nonce pins the exact candidate so a stale request can't release a newer one.

The service refuses with the new `409 GUARDIAN_CANDIDATE_LANDED` when on-chain already matches the candidate's expected post-state (the tx landed; the worker will canonicalize it), and fails closed on an on-chain read error. Otherwise it deletes the candidate delta + matching proposal and clears `has_pending_candidate`, all as hard errors — a 200 means the account is released. `delete_delta` is the linearization point: a `delta_not_found` on retry after a 5xx means the abandon already succeeded.

Also included:

- `remove_candidate` extracted to `jobs/canonicalization/removal.rs` with `RemovalMode { BestEffort, Strict }`; worker behavior unchanged.
- Worker self-heal for a stale `has_pending_candidate` flag (deferred from the #312 review).
- Metric `candidate_outcome="abandoned"`; gRPC label allowlist entry; OpenAPI regenerated.
- Client methods: `GuardianClient::abandon_candidate`, `MultisigClient::abandon_candidate(nonce)`, and TS `abandonCandidate(accountId, nonce)`.

## Summary by CodeRabbit

* **New Features**
  * Added support for abandoning pending canonicalization candidates through the multisig client and Guardian HTTP/gRPC APIs.
  * Added the `POST /delta/candidate/abandon` endpoint, accepting an account ID and nonce and returning abandonment details.
  * Added client library support with typed responses and consistent error handling.
* **Bug Fixes**
  * Prevented abandonment when a candidate has already landed on-chain.
  * Automatically clears stale pending-candidate status when no candidate remains.
* **Documentation**
  * Updated OpenAPI specifications for the new endpoint, request, response, and error conditions.


crates/miden-multisig-client/src/client/proposals.rs
```diff
@@ -556,6 +556,39 @@ impl MultisigClient {
...
+    /// `nonce` pins the exact candidate to release; it is the nonce the
+    /// proposal was pushed with (committed account nonce + 1).
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the candidate's transaction actually landed
+    /// on-chain (`GUARDIAN_CANDIDATE_LANDED` — the server will canonicalize
+    /// it shortly), if no candidate exists at this nonce, or if GUARDIAN
+    /// cannot be reached.
+    pub async fn abandon_candidate(&mut self, nonce: u64) -> Result<()> {
```
zeljkoX: It looks like the TS multisig client hasn't been extended with this new method yet, which breaks SDK parity.

zeljkoX (commented):

I am proposing changes and slightly different architecture. Main idea is that users should send intent to abandon some candidate but resolution should be left to the canonicalization worker.

Details:

# Abandon-as-Intent Architecture

Extend `DeltaStatus::Candidate` with:

- `abandon_requested_at: Option<String>`
- `abandon_confirm_count: u32`

Use `#[serde(default)]` for compatibility. The status remains `candidate`, so the account stays locked until the worker resolves it.

## Endpoint

`POST /delta/candidate/abandon`:

1. Authenticates and validates the candidate nonce.
2. Returns `409 GUARDIAN_CANDIDATE_LANDED` if already landed.
3. Atomically records abandonment intent without overwriting worker counters.
4. Returns `202 Accepted`.

Retries must preserve the original request time and confirmation count. RPC failure should not prevent marking intent because the operation is non-destructive.

## Worker

The canonicalization worker remains the sole lifecycle owner:

- Expected commitment → canonicalize.
- Previous commitment → increment abandonment confirmations.
- Previous commitment after the quarantine duration and required checks → abandon.
- Divergent commitment → reset abandonment confirmations and use existing divergence handling.
- RPC failure → retain and retry.

Suggested defaults:

- `abandon_quarantine_seconds = 30`
- `abandon_quarantine_checks = 2`

## Terminal State

Preserve the delta as history instead of deleting it:

```rust
Discarded {
    timestamp: String,
    reason: DiscardReason::ClientAbandoned,
}
````

Delete/finalize the matching proposal first, transition the delta to discarded, then clear has_pending_candidate. Retry cleanup failures on the next worker run.

## Clients

Rust and TypeScript clients should expose abandonment and polling helpers:
Candidate with intent → waiting
Canonical → landed
Discarded as client-abandoned → successfully abandoned
Missing delta → unexpected


The quarantine reduces late-landing risk but cannot eliminate it without a transaction expiration boundary.

File: docs/openapi.json [hunk 2]
```diff
@@ -1582,6 +1582,105 @@
         ]
       }
     },
+    "/delta/candidate/abandon": {
```
zeljkoX, line 1615: in addition to openapi files let's extend `docs` and `spec` dir where needed. Also README.md files for each crate/package should have updated docs with new methods.

haseebrabbani:
@zeljkoX agreed — this is a better architecture, will adjust the PR to it. Two points to confirm:

1. **Grace-period precedence**: `handle_unverified_candidate` currently defers any unverified candidate inside `submission_grace_period_seconds` without consuming retries. I'll make the abandon-resolution branch take precedence over that deferral (quarantine timing only), otherwise an abandon would still wait out the grace window — flagging in case you intended otherwise.
2. **Scope of `Discarded { reason }`**: should the existing max-retries and divergence discards (which currently hard-delete the delta) also move to reason-carrying `Discarded` in this PR, or keep deletion there and introduce the status only for `ClientAbandoned` for now?

Also planning: intent write preserves `retry_count`/`divergence_count` and is idempotent on retry (original `abandon_requested_at` kept); a divergent observation resets `abandon_confirm_count` and defers to the existing divergence handling; endpoint returns 202 with the current candidate state so clients can poll.

haseebrabbani:
@zeljkoX I reworked to the abandon-as-intent architecture. The intent write is a new unfenced-but-conditional `request_candidate_abandon` storage op that touches only `abandon_requested_at` — worker counters are never overwritten.

Implemented per your spec:
- `DeltaStatus::Candidate` gains `abandon_requested_at` / `abandon_confirm_count` (`#[serde(default)]`, so pre-existing candidates keep deserializing); the status stays `candidate` and the account stays locked until the worker resolves it.
- Endpoint returns **202** with `{state: "pending" | "abandoned", abandon_requested_at}`; still 409 `GUARDIAN_CANDIDATE_LANDED` when already landed; retries are idempotent and preserve the original request timestamp; an RPC failure no longer blocks the request (the intent is non-destructive and the worker re-verifies before finalizing).
- Worker: expected commitment → canonicalize (landed always wins, intent or not); at-base → count a confirmation; divergent → reset the confirmation streak and use the existing divergence handling; RPC failure → retain and retry. Defaults `abandon_quarantine_seconds = 30`, `abandon_quarantine_checks = 2`.
- Terminal state: proposal finalized first, then `Discarded { reason: ClientAbandoned }` (preserved as history), then the pending-candidate flag clears; cleanup failures retry on the next run.
- Clients: all four (Rust `GuardianClient` + `MultisigClient`, TS `guardian-client` + `miden-multisig-client`) expose the abandon request plus a polling helper classifying `waiting` / `landed` / `abandoned` / `unexpected`, via a new additive `discard_reason` field on the proto `DeltaStatus`.
- Docs: `spec/api.md`, `spec/processes.md` (worker behavior), `CONCEPTS.md` failure table, `MULTISIG_SDK.md` API tables, READMEs for all four client crates/packages, OpenAPI regenerated.

Two judgment calls to confirm:

1. **Grace precedence**: the abandon quarantine bypasses the submission-grace deferral (otherwise an abandon would wait out most of the 600s grace). Pinned by `test_abandon_intent_confirms_and_bypasses_grace`.
2. **`Discarded { reason }` scope**: only `ClientAbandoned`, written solely by the abandon finalizer; the worker's retry/divergence discards keep their delete behavior.

Note on TS parity: `packages/miden-multisig-client` depends on the registry `guardian-client@^0.15.0`, so its new methods typecheck against the *next* published guardian-client (verified locally against the built dist).

crates/server/src/builder/canonicalization.rs
```diff
@@ -30,6 +43,8 @@ impl Default for CanonicalizationConfig {
             max_retries: 18,                      // 18 attempts (total: ~3 minutes)
             submission_grace_period_seconds: 600, // Allow proving/submission to settle first
             divergence_confirmations: 2,          // Two ticks to rule out a stale read
+            abandon_quarantine_seconds: 30,       // Let a late-landing tx surface first
```
zeljkoX: What about lowering a bit, 15s? So it does not affect UX too much...

zeljkoX (approved):
LGTM

Thanks for hard work here!

Left nit comment and seems like test are failing.

haseebrabbani:
Thanks! Nit applied in 8c553df (`abandon_quarantine_seconds` 30 → 15, spec updated). That commit also fixes a race found in a follow-up review pass: worker counter writes overwrote the full status blob from a tick-start snapshot and could wipe a concurrently recorded `abandon_requested_at` — `update_candidate_status` now merges the stored intent under the same row lock / delta write lock that `request_candidate_abandon` takes (regression-tested at the filesystem layer).

zeljkoX (approved):
LGTM

Let's resolve conflicts and merge this one.

PR #330, opened 2026-07-20
feat: Remove silent devnet network defaults

### LINKED ISSUE #324: Guardian defaults its RPC endpoint to devnet
(discovered by @0xnullifier) `miden-multisig-client` silently defaults `midenRpcEndpoint` to `https://rpc.devnet.miden.io` when the field is omitted, regardless of the network used by the injected `MidenClient`. 

This means a user can create its main client for testnet while the multisig client creates a second `WasmWebClient` against devnet for proposal execution, FPI/datastore fetches, and commitment checks. That shadow client also reuses the main client’s `storeIdentifier()`, so devnet responses may be processed against testnet state in the same IndexedDB database.

The endpoint should either be required or fail visibly when omitted instead of defaulting to devnet (or anything else). Maybe, it would be better to derive it from the injected `MidenClient` so the two cannot diverge. The guardian server should likely be made stricter as well, since a missing or invalid `GUARDIAN_NETWORK_TYPE` currently falls back to `MidenDevnet`.

zeljkoX:
Require explicit Miden RPC configuration in the TypeScript multisig SDK and GUARDIAN_NETWORK_TYPE at server startup. Fail fast on missing or invalid values, update public helper types, tests, examples, and documentation.

## Summary by CodeRabbit

* **New Features**
  * Server startup now requires a valid `GUARDIAN_NETWORK_TYPE`; supported values are handled case-insensitively.
  * Multisig SDK configuration now requires both Guardian and Miden RPC endpoints.
  * Transaction APIs provide clearer type guidance for required endpoint and signing options.

* **Bug Fixes**
  * Prevented silent fallback to default networks or RPC endpoints when configuration is missing or invalid.

* **Documentation**
  * Updated setup, deployment, Docker Compose, examples, and troubleshooting guidance to reflect required configuration and network-specific endpoints.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,14 +81,19 @@ For env-driven benchmark network/canonicalization settings, apply the runtime co
 #### Running with Cargo
 
 ```bash
-cargo run --bin server
+GUARDIAN_NETWORK_TYPE=MidenTestnet cargo run --bin server
 ```
+
+`GUARDIAN_NETWORK_TYPE` (`MidenLocal`, `MidenTestnet`, or `MidenDevnet`) is
+required — the server refuses to start without it. A root `.env` file works
+too; see [`docs/LOCAL_DEV.md`](docs/LOCAL_DEV.md#environment-file).
 
 EVM proposal support is feature-gated. Default builds do not register EVM
 routes. EVM-enabled servers use the domain-separated `/evm/auth/*`,
 `/evm/accounts`, and `/evm/proposals*` routes.
 
 ```bash
+GUARDIAN_NETWORK_TYPE=MidenTestnet \
 GUARDIAN_EVM_RPC_URLS=31337=http://127.0.0.1:8545 \
 GUARDIAN_EVM_ENTRYPOINT_ADDRESS=0x... \
 cargo run -p guardian-server --features evm --bin server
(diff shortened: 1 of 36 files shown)


haseebrabbani (approved):
lgtm, one comment:

**Silent devnet→testnet switch for existing Docker Compose deployments**

On `main`, `docker-compose.yml` sets no `GUARDIAN_NETWORK_TYPE`, so compose deployments currently run against the server's old `MidenDevnet` default. This PR pins:

```yaml
- GUARDIAN_NETWORK_TYPE=${GUARDIAN_NETWORK_TYPE:-MidenTestnet}
```

which means an existing compose user who pulls and restarts silently moves from devnet to testnet — with their existing volumes/state intact.

A couple of options:

1. Default the compose var to `MidenDevnet` for continuity with current behavior, or
2. Drop the `:-` fallback entirely so compose fails fast like everything else (and have the quickstart docs tell users to set it explicitly).

zeljkoX:
> lgtm, one comment:
> 
> **Silent devnet→testnet switch for existing Docker Compose deployments**
> 
> On `main`, `docker-compose.yml` sets no `GUARDIAN_NETWORK_TYPE`, so compose deployments currently run against the server's old `MidenDevnet` default. This PR pins:
> 
> ```yaml
> - GUARDIAN_NETWORK_TYPE=${GUARDIAN_NETWORK_TYPE:-MidenTestnet}
> ```
> 
> which means an existing compose user who pulls and restarts silently moves from devnet to testnet — with their existing volumes/state intact.
> 
> A couple of options:
> 
> 1. Default the compose var to `MidenDevnet` for continuity with current behavior, or
> 2. Drop the `:-` fallback entirely so compose fails fast like everything else (and have the quickstart docs tell users to set it explicitly).

Thanks for the feedback.

Went with option 2.

Issue #229, opened 2026-05-13
consume_notes proposal verification depends on the caller's local Miden IndexedDB — should it?


VAIBHAVJINDAL3012:
## Summary

Multisig.verifyProposalMetadataBinding works purely from metadata for every proposal type except consume_notes. For consume_notes, it reads the actual Note records from the caller's local Miden IndexedDB and throws if any are missing. This creates a hidden per-device dependency that's surprisingly easy to trip: a legitimate proposal can become permanently un-verifiable for a cosigner whose local store doesn't have the referenced notes, even though the proposal itself is valid and the underlying notes still exist on chain / on the transport.

---

## Versions

- `@openzeppelin/miden-multisig-client@0.14.3`
- `@miden-sdk/miden-sdk@0.14.3`

---

## The dependency

`buildTransactionRequestFromMetadata` dispatches per `proposalType`. All branches except `consume_notes` build the request from metadata alone — `p2id` uses `recipientId`/`faucetId`/`amount`, `add_signer`/`remove_signer`/`change_threshold` use `targetThreshold`/`targetSignerCommitments`, `switch_guardian` uses `newGuardianPubkey`, etc.

`consume_notes` is different. The metadata only carries note IDs:

```ts
// types/proposal.d.ts
export interface ConsumeNotesProposalMetadata extends BaseProposalMetadata {
    proposalType: 'consume_notes';
    noteIds: string[];
}
```

To rebuild the request, the SDK has to fetch the actual `Note` objects (`transaction/consumeNotes.js:5-19`):

```ts
export async function buildConsumeNotesTransactionRequest(client, noteIds, options = {}) {
    if (noteIds.length === 0) throw new Error('At least one note ID is required');

    const rawClient = await getRawMidenClient(client, options.midenRpcEndpoint);

    const noteAndArgsArray = new NoteAndArgsArray();

    for (const noteIdHex of noteIds) {
        const inputNoteRecord = await rawClient.getInputNote(noteIdHex);

        if (!inputNoteRecord) {
            throw new Error(`Note not found in local store: ${noteIdHex}`);
        }

        ...
    }

    ...
}
```

`rawClient.getInputNote(...)` reads from local Miden IndexedDB. No network fallback, no chain query.

---

## Why this is brittle in practice

Whether a cosigner has the referenced note locally depends on factors largely outside the proposal flow.

### Public notes

[...]

PR #338, opened 2026-07-21
fix: propagate P2ID note type so proposals honor private notes (#322)

### LINKED ISSUE #322: P2ID: Guardian proposals always create public notes — noteType not propagated
### Summary
When creating a P2ID proposal via Guardian, the resulting note is always `NoteType.Public`, even if the user selected "Private".

### Details
buildP2idTransactionRequest` hardcodes `NoteType.Public`. No API to override it, and `P2idProposalMetadata` has no field for `noteType`. Present on `upstream/main`.

TS and Rust client implementation should be matched.

haseebrabbani:
## Summary

Fixes #322 — Guardian P2ID proposals always created `NoteType.Public` notes, ignoring the user's "Private" selection. `buildP2idTransactionRequest` (TS) and `build_p2id_transaction_request` (Rust) hardcoded the note type, and proposal metadata had no field to carry it. This PR threads the note type end to end through both clients, matching the TS and Rust implementations.

### Wire format
- New optional `metadata.note_type` field: `"public"` | `"private"`.
- **Omitted when public**, so public payloads stay byte-identical to the pre-#322 wire shape (same precedent as the `consume_notes` v1 handling in #229). Absent ⇒ public, so existing proposals verify and execute unchanged.
- Unknown values are **rejected at parse time** on both sides — the alternative is silently rebuilding a public note that can never match the signed `tx_summary` commitment, which would surface as a confusing "metadata does not match tx_summary" error.

### TS client (`@openzeppelin/miden-multisig-client`)
- `buildP2idTransactionRequest(...)` accepts `options.noteType` (defaults to `NoteType.Public`).
- `createP2idProposal(recipient, faucet, amount, nonce?, { noteType })` records the choice in proposal metadata.
- The rebuild-from-metadata path (cosigner verification + execution) parses `noteType` back, so private proposals are rebuilt as private notes.
- New exports: `parseP2idNoteType`, `p2idNoteTypeToMetadata`, `P2idTransactionOptions`, `P2idNoteVisibility`, `isP2idNoteVisibility`.

### Rust client (`miden-multisig-client`)
- `build_p2id_transaction_request` takes a `NoteType` parameter.
- `TransactionType::P2ID` gains a `note_type` field; `TransactionType::transfer()` still defaults to public, and `transfer_with_note_type()` is the explicit form.
- Threaded through `ProposalBuilder`, execution, `ProposalMetadata`/`ProposalMetadataPayload`, and the export/import round-trip.

### Server + operator client
- `note_type` mirrored in the delta-summary `ProposalMetadata` (and `guardian-operator-client`), so dashboards can surface visibility.
- OpenAPI specs regenerated (`cargo run --features evm --bin gen-openapi -- docs`).

### Examples
- Demo CLI prompts for note visibility on P2ID proposals.
- Browser example wrappers and the smoke harness pass `noteType` through.

## Breaking changes

- Rust: external crates that construct or exhaustively match `TransactionType::P2ID`, or call `build_p2id_transaction_request` / `with_payment_metadata`, need to add the note-type argument. `TransactionType::transfer()` is unchanged.
- TS: additive only.

## Release note

`@openzeppelin/guardian-client` must be published before `@openzeppelin/miden-multisig-client`, since the multisig client's metadata codec references the new wire field.

## Testing

- New Rust tests: default-to-public for absent `note_type`, private threading, unknown-value rejection, public-payload wire-shape stability, and a request-level test that note type changes the built transaction.
- New TS tests: builder default + threading, parse/serialize helpers, codec round-trip + rejection, and an end-to-end `createP2idProposal` test asserting `note_type: "private"` reaches the pushed payload and the rebuild path.
- Full Rust workspace lib tests, clippy, fmt, and `gen-openapi --check` pass; all three TS package suites pass; `tsc` clean for `examples/web` and `examples/smoke-web`.

## Summary by CodeRabbit

* **New Features**
  * Added support for selecting public or private note visibility when creating P2ID proposals.
  * Preserved the selected visibility when proposals are exported, rebuilt, and executed.
  * Added optional note visibility support to browser, web, demo, and smoke-test proposal flows.
  * Extended proposal metadata and API schemas with the optional `noteType` field; omitted values continue to default to public.

* **Bug Fixes**
  * Invalid note visibility values are now rejected with validation errors.
  * Existing proposals without note visibility remain compatible.

### DIFF
--- a/crates/miden-multisig-client/README.md
+++ b/crates/miden-multisig-client/README.md
@@ -63,6 +63,12 @@ let recipient = AccountId::from_hex("0x7b7b7b7a7b7b7b017b7b7b7b7b7b7b")?;
 let faucet = AccountId::from_hex("0x7c7c7c7c7c7c7c017c7c7c7c7c7c7c")?;
 let tx = TransactionType::transfer(recipient, faucet, 1_000);
 
+// P2ID notes are public by default. For a private note (only its hash is
+// published on chain; the recipient needs the note shared out-of-band) use
+// `transfer_with_note_type`. `NoteType` is re-exported from `miden_protocol::note`.
+use miden_protocol::note::NoteType;
+let tx = TransactionType::transfer_with_note_type(recipient, faucet, 1_000, NoteType::Private);
+
 // Proposer creates the delta on GUARDIAN
 let proposal = client.propose_transaction(tx).await?;
 
@@ -152,11 +158,12 @@ on-chain transaction — the integration owns that recipe and drives execution.
 ```rust
 use miden_client::Serializable;
 use miden_multisig_client::{build_p2id_transaction_request, generate_salt};
+use miden_protocol::note::NoteType;
 
 // Producer: build a transaction and propose it under a custom label.
 let salt = generate_salt();
 let mut request = build_p2id_transaction_request(
-    account.inner(), recipient, vec![asset], salt, std::iter::empty(),
+    account.inner(), recipient, vec![asset], NoteType::Public, salt, std::iter::empty(),
 )?;
 let proposal = client.propose_custom_transaction(&request.to_bytes(), "b2agg").await?;
 
--- a/crates/miden-multisig-client/src/account.rs
+++ b/crates/miden-multisig-client/src/account.rs
@@ -339,6 +339,7 @@ mod tests {
                     recipient: account_id,
                     faucet_id: account_id,
                     amount: 10,
+                    note_type: miden_protocol::note::NoteType::Public,
                 })
                 .expect("threshold"),
             1
--- a/crates/miden-multisig-client/src/execution.rs
+++ b/crates/miden-multisig-client/src/execution.rs
@@ -139,6 +139,7 @@ pub async fn build_final_transaction_request(
             recipient,
             faucet_id,
             amount,
+            note_type,
         } => {
             let asset = build_transfer_asset(account, *faucet_id, *amount)?;
 
@@ -146,6 +147,7 @@ pub async fn build_final_transaction_request(
                 account,
                 *recipient,
                 vec![asset.into()],
+                *note_type,
                 salt,
                 signature_advice,
             )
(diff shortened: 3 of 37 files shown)


zeljkoX (approved):
LGTM

I pushed a small change, mostly doc updates made while testing this one.

Review and modify if needed....

I was able to send both note types.

However I was not able to consume private note. Have you tested that? I wonder should we extend this part for clients(if needed) and for testing(demo example)?

haseebrabbani:
@zeljkoX thanks for the doc updates and the `toGuardian` canonicalization — both look good, and the codec tests pass with the change.

On consuming private notes: what you hit is expected with the current SDK surface, not something this PR regresses. A private note only publishes its commitment on chain, so the recipient's Miden client never learns the note's contents via sync — `getConsumableNotes()` can't see it, and `createConsumeNotesProposal` can't fetch it from the local store (FR-012 assumes the proposer has the note). The consume pipeline itself is note-type-agnostic: once the note is in the proposer's store, v2 metadata embeds the full notes, so cosigner verification and execution work unchanged.

The missing piece is the out-of-band note transfer:
1. **Sender** (after executing the private P2ID) exports the created note — the web SDK already has `exportNoteFile(noteId, format)` → `NoteFile`, and the Rust miden-client has the equivalent.
2. **Recipient** imports it (`importNoteFile` / `import_notes`) and syncs; the note then shows up as consumable and the normal `consume_notes` flow applies.

Neither the multisig SDK nor the demo currently wraps export/import, so there's no way to do this without dropping to the raw client. I'd keep this PR scoped to note-type propagation (your docs already flag the out-of-band requirement) and file a follow-up for:

- SDK wrappers: `exportNote(noteId)` / `importNote(noteFile)` on the TS `Multisig` client and Rust `MultisigClient`
- Demo: offer note export after executing a private P2ID, and an "import note file" option in the consume flow
- A demo/smoke scenario covering private P2ID → export → import → consume

Happy to open that issue if you agree.

zeljkoX:
> 1. exportNoteFile

Sounds good.
Let's proceed with this one as is and open an issue.

PR #304, opened 2026-06-25
009 human readable errors impl

haseebrabbani:
## Summary

Implements issue #179 — wallet users now see short, sanitized, human-readable error messages instead of raw backend/transport errors. Implements the error-contract reshape agreed in the spec review (#292): a single `{ code, message, meta }` object, identical on the HTTP body and the gRPC `Status.details`, with the diagnostic `Display` string logged server-side only.

Spec: `speckit/features/009-human-readable-errors/` (`spec.md`, `plan.md`, `tasks.md`).

## Wire contract

```json
{ "code": "authorization_failed", "message": "You're not an authorized signer for this account.", "meta": { "retryable": false } }
```

- `code` — stable machine code (existing vocabulary, unchanged). Branch + i18n key.
- `message` — short, user-safe sentence; always present; safe to display. Wording not stable.
- `meta` — `retryable` (always) + `retry_after_secs` / `missing_permissions` / `paused_at` / `paused_reason` when they apply.
- **HTTP**: body is this object (no `success`, no `error`). **gRPC**: `Status.code` unchanged, `Status.message` = `message`, `Status.details` = this object, for every error.

## What changed (bottom-up)

- **Server** (`crates/server`): `GuardianError::user_message()` (sanitized per-variant, connectivity-vs-internal split) + `retryable()`; reshaped HTTP envelope; `Display` is logged, not returned; gRPC errors now ride on `tonic::Status` (all 9 handlers); rate-limit + `configure` use the user-safe message; OpenAPI `ApiErrorResponse`/`ApiErrorMeta` regenerated.
- **Rust clients**: `ClientError::{guardian_code, user_message, guardian_meta, is_not_found}` parsed from `Status.details`; multisig `get_deltas` branches on `is_not_found()` (not message text).
- **TS `guardian-client`**: `GuardianHttpError` parses `{ code, message, meta }`; replay-retry keys on the `authentication_failed` code.
- **TS `guardian-operator-client`**: parsers + types reshaped to `{ code, message, meta }`.
- **TS `miden-multisig-client`**: `toUserFacingError()` / `isLikelyNetworkError()` connectivity classifier for codeless transport failures (User Story 3 — the 0xMiden/wallet `connectivity-classify.ts` pattern).
- **Examples**: `operator-smoke-web` reads `data.message`.

## Testing

[...]

PR #339, opened 2026-07-21
feat(ts): compiler-checked Guardian error codes (#318)

### LINKED ISSUE #318: guardian-client: `GuardianHttpError.code` is untyped (`string`), so server error codes can't be safely matched
## Summary

The server defines a complete, stable, machine-readable error-code vocabulary, and `guardian-client` throws almost all of it away at the TypeScript boundary. `GuardianHttpError.code` is typed `string | null`, so consumers cannot exhaustively or safely branch on it — typos compile, and new server codes silently fall through.

`guardian-operator-client` already solved this problem with a typed union. `guardian-client` never adopted the pattern.

## Current state

**Server — well defined.** `GuardianError::code()` (`crates/server/src/error.rs:219`) maps every variant to a stable code string, and the JSON envelope (`crates/server/src/error.rs:399`) carries `code`, `error`, plus per-variant fields (`retry_after_secs`, `retryable`, `missing_permissions`, `paused_at`, `paused_reason`, `released_at`).

**Client — untyped.** `packages/guardian-client/src/http.ts:51`:

```ts
public readonly code: string | null;
```

There is no `GuardianErrorCode` type anywhere in the package. The client also lifts only **two of the seven** envelope fields (`code` and `released_at`, `http.ts:67-79`); `retryAfterSecs`, `retryable`, `pausedAt`, and `pausedReason` are left unparsed inside `body`.

## Why this matters

**Three different 409s need opposite handling.** `POST /delta` can return `conflict_pending_delta` (transient — retry), `GUARDIAN_ACCOUNT_PAUSED` (terminal), or `GUARDIAN_ACCOUNT_RELEASED` (terminal, must re-onboard). A consumer that branches on `err.status === 409` will retry-loop a released account forever. Branching on `err.code` is the only correct approach — but the compiler doesn't check it, so a typo silently becomes a no-op.

**Consumers are already string-matching to compensate.** The client itself does this (`http.ts:355`):

```ts
if (retries > 0 && err instanceof GuardianHttpError && err.body.includes('Replay attack')) {
```

Downstream apps do the same, and it misfires. Because `GuardianHttpError.message` is the raw JSON blob, a classifier like `message.includes('commitment') && message.includes('match')` matches `"code":"commitment_mismatch"` (`mismatch` contains `match`) — so a *"your state is stale, rebuild the delta"* signal gets misread as a transient condition and retried.

## Precedent in this repo (and a bug in it)

`packages/guardian-operator-client/src/types.ts:208` defines a real union:

```ts
export type DashboardErrorCode =
  | 'authentication_failed'
  | 'account_not_found'
  // …
```

and normalizes the SCREAMING_SNAKE wire codes to snake_case at the boundary (`http.ts:112`). That's the right shape.

**But it's defeated by widening** (`http.ts:132`):

```ts
code: DashboardErrorCode | string | null;
```

TypeScript collapses `DashboardErrorCode | string` to plain `string`. The literals contribute nothing to the compiler, so there is still no autocomplete, no typo safety, and no exhaustiveness at the access site — the union is documentation-only. This bug should be fixed alongside.

Net: **none of the TS packages currently give a compiler-checked error code.**

## Proposed fix

1. Add `GuardianErrorCode` to `guardian-client`, covering the full set enumerated in `GuardianError::code()` (`crates/server/src/error.rs:219`), using the operator client's snake_case normalization for the three SCREAMING_SNAKE wire codes.
2. Surface it **without** `| string` widening. Either `GuardianErrorCode | (string & {}) | null` to preserve autocomplete, or keep `code: GuardianErrorCode | null` strict and put unrecognized wire codes in a separate `rawCode: string` field so a new server code reads as `null` rather than silently satisfying a `string` branch.
3. Parse the full envelope: `retryAfterSecs`, `retryable`, `pausedAt`, `pausedReason`. Use the envelope's `error` field as the exception `message` instead of the raw blob.
4. Fix the same widening in `guardian-operator-client`.
5. Add a drift guard: a test asserting the TS union matches `GuardianError::code()`.
6. Once codes are typed, replace the `'Replay attack'` substring match in `http.ts:355`.

Note the OpenAPI spec is **not** a viable source of truth here: `ApiErrorResponse` (`crates/server/src/openapi.rs:27`) is a hand-maintained mirror of the private serialized struct, and it declares `code` as bare `"type": "string"` with no enum — a generated client would produce exactly the `string` we already have.

## Blocking prerequisites (server-side)

Two paths mean a typed union would still lie to callers:

- **Rate limiting emits no `code` at all.** The tower middleware (`crates/server/src/middleware/rate_limit.rs:241`) serves 429s for the client API with its own body — `{success, error, retry_after_secs}`, no `code`. So `err.code === 'rate_limit_exceeded'` (the code the enum defines and OpenAPI documents) **never matches**; only `status === 429` works. `GuardianError::RateLimitExceeded` is only constructed on the dashboard path (`crates/server/src/dashboard/util.rs:24`).
- **`/configure` uses a divergent envelope.** It returns `ConfigureResponse` (`crates/server/src/api/http.rs:105`) on both success and failure, with the message in `message` (not `error`) and `code` as `Option`.

Also worth noting for the Rust client: `From<GuardianError> for tonic::Status` (`crates/server/src/error.rs:516`) drops `code` entirely in the `_` arm — only `AccountPaused` and `AccountReleased` attach it via `Status::details`. So `ClientError` (`crates/client/src/error.rs`) can only substring-match `Display` text. That's a separate issue but the same root cause.

haseebrabbani:
> [!IMPORTANT]
> **Release-order blocker:** `@openzeppelin/guardian-client` must be published to npm **before** `@openzeppelin/miden-multisig-client` — the multisig client's new re-exports only resolve against the published guardian-client. Rides the same breaking-changes release window as #304.

## Summary

Implements #318 — the server's stable error-code vocabulary is now compiler-checked in the TS clients instead of collapsing to `string`. Follow-up to #304 (feature 009), tracked as T017 in `speckit/features/009-human-readable-errors/tasks.md`.

### `guardian-client` (new typed surface)

- **`error-codes.ts`**: `GUARDIAN_ERROR_CODES` — the full 38-code vocabulary mirroring `GuardianError::code()` (`crates/server/src/error.rs`), with the three SCREAMING_SNAKE wire forms (`GUARDIAN_ACCOUNT_PAUSED`, `GUARDIAN_ACCOUNT_RELEASED`, `GUARDIAN_INSUFFICIENT_OPERATOR_PERMISSION`) normalized to snake_case — plus the `GuardianErrorCode` union, an `isGuardianErrorCode()` guard, and `normalizeGuardianErrorCode()` for the wire boundary.
- **`GuardianHttpError.code` is now strictly `GuardianErrorCode | null`** (the issue's strict option — no `| string` widening). Comparing against a non-member literal is a **type error**, so typos are caught at compile time and `switch` branches can be exhaustiveness-checked. An unknown code from a newer server reads as `code === null` instead of silently satisfying a string branch.
- **New `rawCode: string | null`** carries the verbatim wire code (including unknown ones) for logging/telemetry.
- **Drift guard**: a test parses the `code()` match arms in `crates/server/src/error.rs` and asserts set equality with the TS union — adding a server code without updating the union (or vice versa) fails the suite. (The OpenAPI spec isn't usable as the source of truth: it declares `code` as a bare `string` with no enum.)

### `guardian-operator-client` (widening bug from the issue)

`DashboardErrorCode | string | null` collapses to plain `string`, making the union documentation-only. Replaced with a named `DashboardErrorCodeOrRaw = DashboardErrorCode | (string & {})` on `ParsedErrorBody.code`, `mapDashboardErrorCode()`, and `GuardianOperatorHttpErrorData.code` — restores autocomplete and literal narrowing. This surface keeps the autocomplete-preserving option (not strict) because its documented contract is to forward raw non-dashboard codes verbatim, and existing tests pin that behavior.

### `miden-multisig-client`

Re-exports the typed-code API for wallet consumers.

## ⚠️ Behavior change (within the #304 breaking window)

`GuardianHttpError.code` now returns the **normalized snake_case** form for the three SCREAMING_SNAKE wire codes (`account_paused`, not `GUARDIAN_ACCOUNT_PAUSED`), matching the operator-client's existing normalization. The raw wire form remains available on `rawCode`. Since #304 is unreleased, this lands inside the same planned breaking-changes release window.

## Incidental fix

`guardian-client`'s build was compiling `*.test.ts` into `dist`; it now uses the same `src/**/*.test.ts` exclude as `guardian-operator-client`, so published artifacts no longer ship compiled tests.

## Testing

- `guardian-client`: 56 tests across 4 files (incl. the drift guard and normalization units), `tsc` build + typecheck clean.
- `guardian-operator-client`: 86 tests + typecheck clean.
- `miden-multisig-client`: typecheck + connectivity tests clean (verified against the locally built `guardian-client`).

Note for CI: `miden-multisig-client` resolves the **published** `@openzeppelin/guardian-client`, so its new re-exports typecheck only once guardian-client is published or linked (same cross-package caveat as #304).

## Out of scope

Item 6 of the issue (the `'Replay attack'` substring match) was already fixed in #304 — replay retry branches on `code === 'authentication_failed'`.

Closes #318.


zeljkoX (approved):
LGTM

I guess like for some other PRs we have blocker that we need to release guardian-client first so multisig client can depend on it. Maybe to add it to OR desc if not already.

haseebrabbani:
@zeljkoX  the "Testing" section already notes that `miden-multisig-client` resolves the *published* `@openzeppelin/guardian-client`, but also added a more prominent warning at the top of the description

Issue #316, opened 2026-07-09
Allow for check and force cannonicalization on main thread when the user knows that the transaction is submitted


0xnullifier:
Currently after sending a transaction from wallet when we re-sync with the guardian operator it takes 10-12s to canonicalise and stop giving the  nonce is too low to import error.

As in the wallet we know that we have sent a transaction it has landed on-chain. We could ask the guardian to check canonicalization in the main thread instead of the worker thread canonicalisation approach which takes time.

A potential api could be - `/check-canonicalize` whose business logic is to fetch the on-chain commitment which should match the current delta commitment as the wallet has already sent the transaction and mark the delta as canonicalised.

Dominik1999:
We could tackle this on our side @zeljkoX 

This is the biggest latency add to a Miden tx at the moment

zeljkoX:
> We could tackle this on our side [@zeljkoX](https://github.com/zeljkoX)
> 
> This is the biggest latency add to a Miden tx at the moment

Thanks for starting conversation here.

We already have some perf improvements in the pipeline https://github.com/OpenZeppelin/guardian/pull/326.

With these improvements we could test and lower default value for worker from current 10s. Additionally perf testing will be done after v0.16.x support work so we can do bigger changes then. 

Just to confirm, this is one of the bigger priorities now?

Dominik1999:
Ah nice, I think the bugs are always the biggest priority, like #322 

Not a top priority, let us see what we can do in v016.

WiktorStarczewski:
Data point from a user hitting exactly this on testnet (`@openzeppelin/miden-multisig-client@0.15.0`, `@openzeppelin/guardian-client@^0.15.2`, 1-of-1 ECDSA multisig):

After a normal `executeProposal` (consume_notes) that landed on-chain (local + on-chain both at nonce 1, commitments matching `0x727b57…`), `getState` kept returning nonce 0 and `syncState()` threw `Refusing to overwrite local state: incoming nonce 0 is not greater than local nonce 1`. In their report it did **not** clear after the usual ~10–12s — it stayed at nonce 0 across many retries.

[...]

PR #344, opened 2026-07-22
fix(multisig-client): keep local state when GUARDIAN nonce lags instead of throwing

### LINKED ISSUE #343: multisig-client: syncState() throws and freezes Multisig.account when GUARDIAN nonce lags behind local
## Summary

`Multisig.syncState()` throws — and leaves `Multisig.account` frozen at its previous snapshot — whenever the GUARDIAN's returned nonce is **≤** the local nonce, i.e. whenever the GUARDIAN is *behind* local. That is the normal state during the post-execute canonicalization window (#316) and a permanent state if a candidate is discarded (#312 / #319). The UI reads `Multisig.account`, so it stays stuck on the pre-execute state even though the local account is already correct and on-chain-verifiable.

## Where

`packages/miden-multisig-client/src/multisig.ts`

- `syncState()` seeds `accountForConfigRefresh = localAccount`, but calls `refreshConfigFromAccount(accountForConfigRefresh)` only **after** the overwrite branch.
- Inside that branch, `ensureSafeToOverwriteLocalState()` **throws** when `incomingNonce <= localNonce`.
- The throw unwinds `syncState()` before `refreshConfigFromAccount()` runs, so `this.account` (assigned only inside `refreshConfigFromAccount`) never advances → UI frozen.

## Reproduction

1. 1-of-1 GUARDIAN multisig; receive funds → sign → `executeProposal` (consume_notes / p2id). Tx lands on-chain; local advances to nonce 1.
2. GUARDIAN still reports nonce 0 (candidate not yet canonical — normal for ~10–12s per #316, or permanent if the candidate was discarded).
3. `syncState()` throws `Refusing to overwrite local state: incoming nonce 0 is not greater than local nonce 1 ...`.
4. `Multisig.account` never refreshes → UI frozen at the pre-execute snapshot.

Reported with `@openzeppelin/miden-multisig-client@0.15.0` on testnet; the same code is on `main` (0.15.2).

## Expected

The guard "don't overwrite good local state with a lower-nonce GUARDIAN state" is correct, but a GUARDIAN that is simply *behind* is not an error. When local is ahead (and independently verifiable against chain via `verifyStateCommitment`), `syncState()` should **keep local, refresh config from it, and return** — not throw. Genuine divergence (same nonce / different commitment, or incoming-vs-on-chain mismatch) must still throw.

## Proposed fix

Make the nonce guard a non-fatal signal: a `isSafeToOverwriteLocalState()` helper returns `false` when the GUARDIAN is behind (skip overwrite, keep local) instead of throwing; `syncState()` guards `newAccount()` on that boolean while still running `refreshConfigFromAccount(localAccount)`. Keep the commitment-mismatch and same-nonce-divergence throws.

This is the client-side complement to #316 (which removes the server-side lag): #316 shortens how long the GUARDIAN stays behind; this stops the client freezing during whatever lag remains, and handles the permanent-behind (discarded) case gracefully.

WiktorStarczewski:
## What

`Multisig.syncState()` threw — and left `Multisig.account` frozen at its previous snapshot — whenever the GUARDIAN's nonce was `<=` the local nonce, i.e. whenever the GUARDIAN was simply *behind* local. `refreshConfigFromAccount()` runs only after the overwrite branch, so the throw in `ensureSafeToOverwriteLocalState()` unwound `syncState()` before `this.account` was ever refreshed. The UI (which reads `Multisig.account`) then stayed frozen on the pre-execute state even though the local account was already correct and independently on-chain-verifiable.

This is the normal state during the post-execute canonicalization window (#316) and permanent if the candidate is discarded (#312 / #319).

## Change

- Rename `ensureSafeToOverwriteLocalState` → `isSafeToOverwriteLocalState`, returning `boolean` instead of throwing on the nonce guard.
- When the GUARDIAN is strictly **behind** local (`incomingNonce < localNonce`): return `false` → `syncState()` skips the overwrite, keeps local, and still calls `refreshConfigFromAccount(localAccount)` so `this.account` advances and the UI unfreezes.
- Genuine divergence still throws: same nonce + different commitment, and incoming-vs-on-chain commitment mismatch.

The guard "don't clobber good local state with a lower-nonce GUARDIAN state" is preserved — it is now a *skip*, not a fatal error. This is the client-side complement to #316: #316 removes the server-side lag; this stops the client freezing during whatever lag remains and handles the permanent-behind (discarded) case gracefully.

## Test plan

`packages/miden-multisig-client/src/multisig.test.ts` (86/86 pass):
- Rewrote the "incoming nonce lower than local" test: `syncState()` now resolves, does **not** call `newAccount`, needs no on-chain round-trip, and refreshes config from the local account.
- Updated the equal-nonce / divergent-commitment test: still throws (new message).
- Added a regression test for the exact repro (local nonce 1, GUARDIAN nonce 0 → `Multisig.account` advances to the local nonce-1 account).

`tsc --noEmit` adds no new errors from this change; `multisig.ts` appears in zero type errors. (The pre-existing `guardian-client` `.code` / `GuardianErrorMeta` errors are unrelated — see #318 — and reproduce on `main`.)

Closes #343

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved account synchronization when guardian-provided state is behind the local state.
  * Preserved the local account state and configuration without unnecessary network requests in supported cases.
  * Prevented accounts from remaining frozen after execution when synchronization detects an outdated guardian nonce.
  * Continued to detect and report genuine state conflicts, including commitment mismatches.

### DIFF
--- a/packages/miden-multisig-client/src/multisig.test.ts
+++ b/packages/miden-multisig-client/src/multisig.test.ts
@@ -541,7 +541,7 @@ describe('Multisig', () => {
       expect(mockWebClient.newAccount).not.toHaveBeenCalled();
     });
 
-    it('should throw when incoming state nonce is lower than local nonce', async () => {
+    it('keeps local state and refreshes config from it when GUARDIAN nonce is behind local', async () => {
       const config = {
         threshold: 1,
         signerCommitments: ['0x' + 'a'.repeat(64)],
@@ -558,7 +558,8 @@ describe('Multisig', () => {
         'https://rpc.devnet.miden.io'
       );
 
-      mockWebClient.getAccount.mockResolvedValueOnce(mockedAccount('0x' + 'a'.repeat(64), 3));
+      const localAccount = mockedAccount('0x' + 'a'.repeat(64), 3);
+      mockWebClient.getAccount.mockResolvedValueOnce(localAccount);
       mockAccountDeserialize.mockReturnValueOnce(mockedAccount('0x' + 'b'.repeat(64), 2));
       mockFetch.mockResolvedValueOnce({
         ok: true,
@@ -570,11 +571,24 @@ describe('Multisig', () => {
           updated_at: '2024-01-02T00:00:00Z',
         }),
       });
+      mockDetectConfig.mockReturnValueOnce({
+        threshold: 2,
+        numSigners: 2,
+        signerCommitments: ['0x' + '1'.repeat(64), '0x' + '2'.repeat(64)],
+        guardianEnabled: true,
+        guardianCommitment: '0x' + 'd'.repeat(64),
+        vaultBalances: [],
+        procedureThresholds: new Map(),
+      });
 
-      await expect(multisig.syncState()).rejects.toThrow(
-        'incoming nonce 2 is not greater than local nonce 3'
-      );
+      // GUARDIAN behind local (nonce 2 < 3): no throw, local kept, no overwrite,
+      // and the decision needs no on-chain round-trip.
+      await expect(multisig.syncState()).resolves.toBeDefined();
       expect(mockWebClient.newAccount).not.toHaveBeenCalled();
+      expect(mockRpcGetAccountDetails).not.toHaveBeenCalled();
+      // Config refreshed from the authoritative local account (UI unfreezes).
+      expect(multisig.account).toBe(localAccount);
+      expect(multisig.threshold).toBe(2);
     });
 
     it('should throw when incoming state nonce equals local nonce but commitment differs', async () => {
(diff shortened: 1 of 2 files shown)


zeljkoX (approved):
Thanks for contribution <3.

LGTM

This PR introduces small divergence with Rust client

crates/miden-multisig-client/src/client/account.rs

We should update Rust client logic to match TS client. 

We can open an issue for later if not handled within this PR.

zeljkoX:
@WiktorStarczewski 

Just noticed that merging is blocked due to: "Commits must have verified signatures".

Setting up [GH commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and signing PR commits would be needed unblock PR.

WiktorStarczewski:
> @WiktorStarczewski
> 
> Just noticed that merging is blocked due to: "Commits must have verified signatures".
> 
> Setting up [GH commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and signing PR commits would be needed unblock PR.

Done. Not sure what happened to my signing setup for those commits honestly, will check.

PR #346, opened 2026-07-22
fix: build P2ID asset from vault entry to match callback-flag vault keys

### LINKED ISSUE #336: Handle fungible assets via vault key when building P2ID notes
## Summary

Asset handling for P2ID note creation should be changed to construct the outgoing `FungibleAsset` from the account vault's **vault key** instead of building it directly from the faucet ID. This preserves callback metadata encoded in the vault key, which is lost when the asset is constructed with `new FungibleAsset(faucetId, amount)`.

This mirrors the change made in the multisig client (`buildP2idTransactionRequest` in `packages/miden-multisig-client/src/transaction/p2id.ts` of the private-state-manager repo):

```ts
// Before
const asset = new FungibleAsset(faucet, amount);
const noteAssets = new NoteAssets([asset]);

// After — look the asset up in the sender's vault and rebuild it from its vault key
const asset = account
  .vault()
  .fungibleAssets()
  .find((asset) => asset.faucetId().toString() === faucetId);

if (!asset) throw new Error("Asset Not found in the vault");

const newAsset = FungibleAsset.fromVaultKey(asset.vaultKey(), amount);
const noteAssets = new NoteAssets([newAsset]);
```

The sending account is now required as an input so the asset can be resolved from its vault.

## Dependency

This depends on the new web-sdk helpers `FungibleAsset.fromVaultKey(vaultKey, amount)` and `FungibleAsset.vaultKey()`:

- https://github.com/0xMiden/web-sdk/pull/247

0xnullifier:
## Summary

Closes #336.

Assets with `AssetCallbackFlag::Enabled` encode the flag in their vault key, so an asset built with `new FungibleAsset(faucet, amount)` does not match the vault entry even for the same faucet id. `buildP2idTransactionRequest` now resolves the outgoing asset from the sender's vault (`account.vault().fungibleAssets()`) and rebuilds it via `FungibleAsset.fromVaultKey(vaultKey, amount)`, threading the sending `Account` through the builder and its callers (`Multisig.createP2idProposal`, the metadata rebuild path, and the browser example).

- Bumps `@miden-sdk/miden-sdk` to 0.15.8 for the new `FungibleAsset.fromVaultKey` / `vaultKey()` helpers (0xMiden/web-sdk#247)
- Adds tests pinning the `fromVaultKey` usage and the vault-miss error path

**Why this does not break existing `Multisig` flows**

- **Co-signer verification stays consistent.** `verifyProposalMetadataBinding` rebuilds the request and immediately re-executes it via `executeForSummary` against the web client's store. Any state in which the new vault lookup could fail (asset absent) is state in which that execution already failed before this change — the lookup introduces no new failure mode, it only surfaces the same one earlier with a clearer message.
- **Rebuilt requests remain byte-identical to what was signed.** The vault key for a live balance cannot change representation between proposal creation and verification, so the request reconstructed from metadata still matches the bound `tx_summary` commitment.
- **Pre-existing pending proposals reverify unchanged.** For assets without the callback flag, `FungibleAsset.fromVaultKey(vaultKey, amount)` produces the same bytes as the old `new FungibleAsset(faucet, amount)` construction.
- **State freshness follows the existing contract.** `this.account` is refreshed by `syncState()`, which callers already must invoke after executed transactions (consuming a funding note, etc.) for other flows to work; this change adds no new sync requirement.
- **The public `Multisig` API is unchanged.** Only the exported helper `buildP2idTransactionRequest` gained a parameter, and it should have not have any known consumers

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved multisignature P2ID transaction construction by using the account’s available faucet asset.
  * Added validation and clearer failure handling when the required asset is unavailable.
  * Ensured transaction proposals and rebuilt transactions use consistent account information.

* **Tests**
  * Expanded coverage for matching vault assets and missing-asset scenarios.

* **Chores**
  * Updated the Miden SDK dependency to version 0.15.8 across examples and the multisig client.


haseebrabbani:
@0xnullifier The fix looks right, but the vault lookup and the execution read different state: the lookup uses `this.account` (a snapshot only refreshed by `Multisig.syncState()`), while `executeForSummary` runs against the web client's store, which other flows update directly (e.g. the finalize path calls `webClient.syncState()` without refreshing `this.account`).

Concrete regression window: fund a fresh multisig, finalize the consume-notes proposal, then call `createP2idProposal` without an explicit `multisig.syncState()` — the store has the asset but the snapshot doesn't, so this now throws `Asset not found in vault` where it previously succeeded. Same skew applies to the co-signer rebuild path.

Could you pass the store's account at both `multisig.ts` call sites instead?

​```ts
const account = await webClient.getAccount(AccountId.fromHex(this._accountId));
​```

That makes the lookup source identical to what `executeForSummary` executes against, and the "no new failure mode" argument in the description becomes exactly true.

0xnullifier:
@haseebrabbani addressed your comments

zeljkoX (commented):
@0xnullifier 

The remaining difference between Rust and TS sdk is robustness. Missing faucet asset: Rust defaults to Disabled and continues (execution.rs); TS throws. Is that intended? Should they match?

0xnullifier:
@zeljkoX 

done mirrored how it's handled on rust side if the vault key is not present then we just do a `Disabled` one

zeljkoX (approved):
LGTM

Thanks!

Note: We would soon need this change propagated to miden web sdk 0.16.x

zeljkoX:
@0xnullifier Sorry, could you resolve merge conflicts so we can merge it and include in tomorrow release?
Seems like fix for private notes introduced merge conflicts here.
Thanks!

PR #347, opened 2026-07-23
feat: Add fast track canonicalization check for recent candidates

### LINKED ISSUE #316: Allow for check and force cannonicalization on main thread when the user knows that the transaction is submitted

zeljkoX:
This PR adds a two-tier canonicalization cadence. Alongside the existing full pass,
a lightweight fast-promotion pass probes only recent candidate deltas on a
shorter timer, so freshly-pushed deltas canonicalize sooner without waiting
for the full sweep.

<h3>Performance impact</h3><p>Using the real multisig E2E benchmark against Miden testnet, with two persistent accounts alternating P2ID transfers, fast-track canonicalization improved the comparable 20-operation run:</p><div>

| Metric | Before | Fast track | Improvement |
|---|---:|---:|---:|
| Runtime | 5m51s | 4m14s | 28% faster |
| Proposal p50 | 4.50s | 3.24s | 28% faster |
| Proposal p95 | 10.13s | 5.82s | 43% faster |
| Execution p50 | 5.14s | 2.16s | 58% faster |
| Canonicalization p50 | 5.92s | 4.40s | 26% faster |
| Pending-delta retries | 56 | 38 | 32% fewer |

## Summary by CodeRabbit

* **New Features**
  * Added a fast promotion pass that can promote recently eligible candidates more frequently, while preserving full-pass validation and lifecycle handling.
  * Added configurable fast-promotion intervals and eligibility windows, enabled by default and controllable through deployment settings.
  * Added pagination and cross-storage support for efficient recent-candidate processing.

* **Monitoring**
  * Added metrics for fast-run counts and durations.
  * Expanded startup and troubleshooting information.

* **Documentation**
  * Updated configuration, infrastructure, API, and process documentation for fast promotion behavior.


haseebrabbani (approved):
lgtm, small comment:

`recent_candidate_cursor` builds the cursor from the page's last candidate *before* the page is processed, and a `parse_from_rfc3339` failure propagates with `?`, erroring out the entire `process_recent_pass`:

```rust
let next_cursor = Self::recent_candidate_cursor(last_candidate)?;
```
Since the fast pass runs every 3 seconds, one candidate with an unparseable status.timestamp() means a steady stream of outcome=error fast runs — and it self-heals only once the full pass eventually discards that candidate, which can take minutes given the 600s grace period.

Suggestion: log and skip the malformed candidate (or end the page there and clear the cursor) instead of failing the whole pass, so one bad row can't take out fast promotion entirely.

PR #349, opened 2026-07-23
345 Retain-and-reconcile: auto-recover accounts whose stored state fell behind the chain

### LINKED ISSUE #345: GUARDIAN does not auto-recover an account whose stored state is permanently behind on-chain
## Problem

When a guarded account's canonicalization fails to complete, the GUARDIAN's stored state (`getState`) can be left **permanently behind** the on-chain nonce with **no path back to correct on its own**. Local + on-chain reach nonce N+1, but `getState` stays at nonce N indefinitely. This blocks the account: every new proposal is built on the GUARDIAN's stored base (`push_delta_proposal` uses `prev_commitment = current_state.commitment`), so while the base is stale, new guardian txs mismatch and fail.

Reported on testnet (`@openzeppelin/miden-multisig-client@0.15.0`): after a normal `executeProposal` that landed on-chain (local == on-chain at nonce 1, commitment `0x727b57…`), `getState` stayed at nonce 0 across 10+ minutes and never recovered.

This is distinct from #343 / #344, which fix only the *client symptom* (`syncState()` throwing and freezing the UI). Those do **not** unstick the GUARDIAN's own state.

## Why it can't just self-heal (verified against the code)

The GUARDIAN can only read the on-chain **commitment** (a hash) — never the account state. Guarded accounts are **private** by construction, so only a commitment is on-chain; the full state lives off-chain, supplied by the client.

- The server's only production node call is `get_account_commitment` (returns a 32-byte hash) inside `verify_state` — `crates/server/src/network/miden/mod.rs:75-119`, `crates/miden-rpc-client/src/lib.rs:135-176`. `get_account_details` is dead code and even requests `details: None`.
- State advances via `apply_delta` on the **client-supplied** `state_json` preimage — it never fetches state from chain.
- Guarded accounts default to `AccountType::Private` (`crates/contracts/src/multisig_guardian.rs:61`); the production create path never overrides it (`crates/miden-multisig-client/src/client/account.rs:113-125`).

A hash cannot be inverted, so for a **diverged** private account no server change can reconstruct the correct state — only the client (which holds it) can.

Two more facts that shape the fix:
- **Discard deletes the delta** (canonicalization discard → `delete_delta`; on `main` via `remove_candidate`). Once deleted, nothing remains server-side to re-canonicalize.
- **Monotonicity:** a diverged candidate's `new_commitment` is a sibling that was never committed; the chain committed a *different* state and only moves forward, so "retain the delta and re-check `base + delta == on_chain`" can never converge for a genuine divergence.

## The two cases

| Case | Auto-recoverable? |
|---|---|
| **(a) Valid delta, correct base, discarded** (RPC flaky / worker down / past retry budget) | **Yes — with a server change.** The delta was correct; it just needs re-checking once on-chain catches up. |
| **(b) True divergence** (stored base is actually wrong) | **No, fundamentally** — private account ⇒ only the hash on-chain ⇒ server cannot reconstruct state. Client must re-supply it. |

## Proposed design (two parts → two PRs)

**1. Server — retain-and-reconcile (auto-recovers case (a), zero client calls).**
On discard, instead of deleting the delta, keep it under a **new status** (not `candidate`, so it does **not** re-hold `has_pending_candidate` — the push-time lock scans candidate-status delta rows: `push_delta.rs:53-68`, `storage/postgres.rs:562-581`; this is the #319 lock trap). A bounded background reconcile pass re-checks `stored_base + delta == on_chain` and canonicalizes if/when it matches. TTL-bounded so it can't retry forever. Only helps non-diverged deltas (monotonicity); diverged ones are dropped as today.

**2. Client — reconcile a lagging base in the execute/propose path (handles case (b), no new user-facing method).**
`push_delta_proposal` / `push_delta` already build on and check against the GUARDIAN's stored base (`prev_commitment == current_state.commitment`, `push_delta.rs:70-75`), so a base mismatch is discovered *exactly where it matters*. Reconcile there — re-supply the on-chain-verified local state via `configure`, gated on `verifyStateCommitment()` — then proceed.

**Explicitly rejected — do NOT fold the reconcile write into `syncState()`.** `syncState` is a read an app calls on a timer; turning it into a push would fire during the *normal* canonicalization-lag window and overwrite the base out from under an in-flight delta, stranding/discarding a signed proposal (and racing across cosigners). `configure` is unguarded (no pending-candidate / no chain check — `configure_account.rs`), so gating on `verifyStateCommitment` fixes "push a wrong state" but not "clobber an in-flight delta." Keep the reconcile in the execute path.

## Deeper fix

True divergence means the base became wrong, which shouldn't happen in normal operation — likely concurrent-proposal races (see #303). Fixing the divergence source reduces how often recovery is even needed; auto-recovery is the safety net, not the cure.

## Related
- #343 / #344 — client symptom fix (syncState no longer throws/freezes). Does **not** fix this.
- #316 — client-triggered force-canonicalize for the ~10–12s lag (the fast-path complement to server retain-and-reconcile).
- #319 — stranded-candidate lock / abandon (the lock the retain-status must avoid re-holding).
- #312 — diverged-candidate fast-discard (still deletes the delta).
- #303 — likely divergence source.

_Analysis produced by a multi-agent code investigation over `crates/server` + `packages/miden-multisig-client`; every load-bearing claim adversarially verified against the tree._

haseebrabbani:
Closes #345. Related: #312, #319/#327, #326, #303.

### Root cause

#312's diverged fast-discard shipped in v0.15.1/v0.15.2 without #326's empty-digest shield (merged ~1.5h after #345 was filed). A new account's unlanded first tx reads on-chain as `0x000…0` → misclassified as diverged → candidate deleted ~2 ticks after submission, while still proving. The tx then lands, and since guarded accounts are private (only a hash on-chain), the deleted delta is unrecoverable — stored state permanently stuck behind the chain, exactly the reported incident. `main` already has the shield; this PR removes the deeper hazard: any wrong give-up verdict + delete is unrecoverable.

### Design

Nothing is deleted when the worker gives up. Retry-exhausted and confirmed-diverged candidates become `retained` (labeled `retry_exhausted` / `diverged`):

- The account is released immediately; a new submission at the same nonce supersedes the retained row.
- The canonicalization pass reconciles retained rows (same verification, fenced promotion, only while no candidate is in flight): if the chain ever shows the delta's commitment, the stuck base recovers automatically.
- TTL-bounded (`retained_ttl_seconds`, default 24h; `0` = legacy delete). Genuinely diverged deltas can never match (monotonicity) and age out.
- `discarded { client_abandoned }` rows within the TTL are reconciled too — the late-landing net for the #319 abandon quarantine.

The issue proposed still dropping diverged candidates; this PR retains them too, since #312 proved the diverged verdict itself can be wrong.

### Also fixed en route

- Resubmit-after-abandon wedge (unreleased, #327): the kept abandoned row blocked resubmission at its nonce forever on Postgres.
- Wrong-row promotion race: promotions now gate on the exact lifecycle kind the verification read, so a superseding candidate can't be stamped canonical with the old delta's commitment.
- Filesystem worker/API races (lock now held across lifecycle writes), pending-proposal leak on supersede, abandon-intent wipe on retain, abandon endpoint 404/false-409 on retained rows, and the operator dashboard client throwing on any `retained` feed row.

## Summary by CodeRabbit

* **New Features**
  * Added a `retained` delta status for candidates kept after verification retries are exhausted or commitments diverge.
  * Retained candidates are automatically reconciled in the background and may later become canonical.
  * Dashboard feeds, filters, counts, APIs, and client libraries now support retained deltas and retention reasons.
  * Added configurable retention duration, defaulting to 24 hours; setting it to zero restores prior behavior.

* **Documentation**
  * Updated lifecycle, recovery, and API documentation to describe retained deltas and reconciliation.


crates/server/src/jobs/canonicalization/reconciliation.rs
```diff
@@ -0,0 +1,491 @@
+//! Background reconciliation of recoverable deltas (issue #345).
```
zeljkoX, line 1: processor.rs is already big and could be refactored so moving logic here is good approach. Nice!

zeljkoX (approved):
Approach looks good after using separate worker with different time and backoff. Forwarded few additional items to double check and improve. Other than that great work!

zeljkoX (approved):
LGTM

Thanks!

PR #350, opened 2026-07-23
fix: count only viable proposals toward the pending-proposal cap

### LINKED ISSUE #337: Add cancel/delete API for pending proposals and candidate deltas (multisig.deleteProposal)
## Problem

There is currently **no way to cancel a pending proposal or discard a stuck candidate delta on demand**. If a transaction is proposed and something goes wrong (e.g. proving fails after the delta is pushed), the account is locked until the canonicalization worker gives up:

- `push_delta` and `push_delta_proposal` both reject with `GuardianError::ConflictPendingDelta` (HTTP 409) while `has_pending_candidate` is set — see `crates/server/src/services/push_delta.rs` and `crates/server/src/services/push_delta_proposal.rs`.
- The only cleanup is the canonicalization worker's auto-discard (grace period + retries, ~13 min with defaults in `crates/server/src/builder/canonicalization.rs`), or a manual DB edit.

## Proposal

Add a delete/cancel path exposed end-to-end:

1. **Server**: a `DELETE /delta/proposal` (authenticated) endpoint that deletes the proposal **and** its matching candidate delta if one was pushed, clearing `has_pending_candidate`. All the plumbing already exists — the worker's `remove_candidate` path in `crates/server/src/jobs/canonicalization/processor.rs` already performs the atomic candidate-delete + flag-clear + proposal-delete sequence; only an HTTP handler + thin service wrapper are missing. The EVM side already has exactly this shape: `POST /evm/proposals/{proposal_id}/cancel` in `crates/server/src/api/evm.rs`.
2. **Client**: a `multisig.deleteProposal(commitment)` method in `packages/miden-multisig-client` (and the Rust client) so a proposer can cancel and re-propose without waiting for auto-discard or touching the database.

haseebrabbani:
Closes #337 (phase 1 — the issue was narrowed to stale-proposal capacity after #327 landed the candidate-abandon half).

## Problem

Proposals are only deleted when they win canonicalization or are abandoned along with their candidate (#327). A proposal that loses the race or never reaches threshold stays `Pending` forever, and `push_delta_proposal` counted every pending proposal against `GUARDIAN_MAX_PENDING_PROPOSALS_PER_ACCOUNT` (default 20). Capacity was therefore consumed monotonically: after 20 dead proposals over an account's lifetime, every new proposal permanently failed with `409 PendingProposalsLimit`, and the only remedy was a manual DB edit.

## Fix

The capacity check now counts only *viable* pending proposals — those whose `prev_commitment` matches the current account state commitment. A proposal built on a superseded commitment can never become a candidate, so it no longer consumes capacity. Viability is derived from `prev_commitment` (not nonce alone), per the design discussion in #337.

Notes on scope:

- The filter lives in the shared service layer using the `current_state` already fetched there, so it covers HTTP and gRPC and works identically for the Postgres and filesystem backends. No new API surface, no schema change.
- Non-viable proposals stay in storage and remain visible via `pull_pending_proposals` — they just stop counting against the limit. Hard deletion/retention and the archive/unarchive design remain follow-up work (phase 2).

### DIFF
--- a/crates/server/src/services/push_delta_proposal.rs
+++ b/crates/server/src/services/push_delta_proposal.rs
@@ -104,8 +104,18 @@ pub async fn push_delta_proposal(
             GuardianError::StorageError(format!("Failed to load pending proposals: {e}"))
         })?;
 
+    // Only viable proposals consume capacity. A proposal built on a
+    // superseded commitment can never become a candidate, so counting it
+    // would let dead proposals accumulate until the account is permanently
+    // locked out with PendingProposalsLimit (#337). Non-viable proposals
+    // stay in storage and remain visible via pull_pending_proposals.
+    let viable_pending = pending_proposals
+        .iter()
+        .filter(|record| record.proposal.prev_commitment == current_state.commitment)
+        .count();
+
     let max_pending_proposals = max_pending_proposals_per_account();
-    if pending_proposals.len() >= max_pending_proposals {
+    if viable_pending >= max_pending_proposals {
         return Err(GuardianError::PendingProposalsLimit {
             limit: max_pending_proposals,
         });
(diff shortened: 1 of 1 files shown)


zeljkoX (approved):
LGTM

PR #361, opened 2026-07-30
feat: SDK note export/import and demo flow for out-of-band private note transfer (#356)

### LINKED ISSUE #356: Support consuming private notes: SDK export/import wrappers and demo flow for out-of-band note transfer
Follow-up to #338 (see https://github.com/OpenZeppelin/guardian/pull/338#issuecomment-5049641784), which made proposals honor the P2ID note type (#322). That PR propagates the note type correctly, but a recipient still cannot consume a **private** P2ID note through the multisig SDK or demo.

## Problem

A private note only publishes its commitment on chain, so the recipient's Miden client never learns the note's contents via sync:

- `getConsumableNotes()` can't see the note.
- `createConsumeNotesProposal` can't fetch it from the local store (FR-012 assumes the proposer has the note).

The consume pipeline itself is note-type-agnostic — once the note is in the proposer's store, v2 metadata embeds the full notes, so cosigner verification and execution work unchanged. The missing piece is the **out-of-band note transfer**:

1. **Sender** (after executing the private P2ID) exports the created note — the web SDK already has `exportNoteFile(noteId, format)` → `NoteFile`, and the Rust miden-client has the equivalent.
2. **Recipient** imports it (`importNoteFile` / `import_notes`) and syncs; the note then shows up as consumable and the normal `consume_notes` flow applies.

Neither the multisig SDK nor the demo currently wraps export/import, so there's no way to do this without dropping to the raw client. The docs added in #338 already flag the out-of-band requirement.

## Proposed work

- **SDK wrappers:** `exportNote(noteId)` / `importNote(noteFile)` on the TS `Multisig` client and Rust `MultisigClient`
- **Demo:** offer note export after executing a private P2ID, and an "import note file" option in the consume flow
- **Testing:** a demo/smoke scenario covering private P2ID → export → import → consume

haseebrabbani:
Closes #356. Follow-up to #338, which made proposals honor the P2ID note type but left no way to hand a private note to its recipient without dropping to the raw Miden client.

## What's included

**SDK wrappers**
- Rust `MultisigClient`: `export_note` / `export_note_to_bytes` (full note with proof when the inclusion proof is known, details-only otherwise), `import_note` / `import_note_from_bytes`, and `p2id_note_id(&proposal)` — the deterministic note ID of a P2ID proposal, resolvable ahead of execution.
- TS `Multisig`: `exportNote` / `importNote` over the web SDK's `NoteFile`, plus `getP2idNoteId` built on the shared P2ID note builder (extracted as `buildP2idNoteFromMetadata` so proposal creation and ID recomputation can't drift).

**Demo**
- After executing a private P2ID, the demo announces the created note and offers to export the note file.
- The consume-notes flow gains an import option (menu entry when no notes are cached, `i` shortcut otherwise) that imports and syncs before listing.

**Smoke coverage**
- New `getP2idNoteId` / `exportNote` / `importNote` commands on the smoke-web harness.
- New `private-note-roundtrip-canary` scenario in both workflow matrices: private P2ID → export → import → consume across separate local stores.

## Summary by CodeRabbit

- **New Features**
  - Added private note export and import for offline, out-of-band transfers.
  - Added deterministic P2ID note ID lookup before transaction execution.
  - Added support for importing notes and making them consumable after synchronization.
  - Improved demo flows with prompts to export, import, sync, and consume private notes.

- **Documentation**
  - Added guidance for private note sharing, export/import workflows, and synchronization requirements.
  - Expanded SDK references for the new note management capabilities.

### DIFF
--- a/.agents/skills/smoke-test-rust-multisig-sdk/references/workflow-matrix.md
+++ b/.agents/skills/smoke-test-rust-multisig-sdk/references/workflow-matrix.md
@@ -191,7 +191,7 @@ Steps:
 5. In tab B, sync, then open `[4] Proposal management -> [1] Create proposal -> [4] Consume notes` and confirm the private note is NOT listed (only its commitment is on chain).
 6. In the consume flow, enter `i` (or choose `[2] Import a note file` from the empty-notes menu) and give the exported file's path. The demo imports the note and syncs automatically.
 7. Select the imported note and create the consume-notes proposal in tab B.
-8. In tab A, sign the consume-notes proposal.
+8. In tab A, sync until the private note shows as consumable there too, then sign the consume-notes proposal. (The sender's store knows the full note and self-heals once a sync attaches the inclusion proof; a cosigner tab that never created nor imported the note must import the note file first.)
 9. Execute once signatures are sufficient, then sync both tabs.
 10. Verify the vault balance reflects the reconsumed asset.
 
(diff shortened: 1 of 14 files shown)


zeljkoX (approved):
This looks great. Thanks for working on this.
I have tested it through rust demo and it works great.

I would suggest small changes and matching functionality between sdks.

Atm TS SDK does not expose  exportNoteToBytes/importNotesToBytes like Rust sdk does.

In addition to that I am proposing to change SDK method names from 

importNote -> importNoteFromFile
exportNote -> exportNoteToFile

So in the end both SDK would expose:
- importNoteFromFile
- exportNoteToFile
- exportNoteToBytes
- importNoteFromBytes

PR #362, opened 2026-07-30
feat: extend multisig sdks with prover config and retries logic

### LINKED ISSUE #360: Make multisig SDK prover retries configurable (429 / timeout), default 2 attempts
Both multisig SDKs should expose a configurable retry budget for transaction
proving, applied to transient remote-prover failures (gRPC `Cancelled` /
`"Timeout expired"`, `Unavailable`, HTTP 429). Proposed default: **2 attempts**
(one retry). Today the Rust SDK hardcodes the budget and the TypeScript SDK has
no retry at all.

## Why

Surfaced during the #317 scalability measurement runs against testnet. The
`for_testnet()` / `for_devnet()` presets delegate proving to a shared remote
prover (`tx-prover.testnet.miden.io`), which cancels requests under load:

| Concurrent writers | Proving failure rate |
|---|---|
| 8 | 45% |
| 16 | 45% |

An identical proportion at double the offered load is the signature of a
saturated queue, not a rejected transaction — the same proof succeeds on a
later attempt. A bounded retry took the 16-writer failure rate to **0%**
(243/243 operations, zero stranded candidates) and left ~13% at 64 writers.

The Miden node also advertises `x-ratelimit-limit: 128`; at 32 concurrent
writers 437 requests were rejected with 429. Reads already back off on 429 in
the benchmark harness, but nothing in the SDKs does.

## Scope

- **Rust** (`crates/miden-multisig-client`): 
- **TypeScript** (`packages/miden-multisig-client`)

## Requirements

1. Retry count is configurable per client; the budget counts total attempts, so
   `1` disables retrying and `0` is coerced to `1`.
2. Default **2 attempts**. See the note below — this differs from what was
   measured.
3. Backoff is exponential with jitter and a hard ceiling. The cap must be
   applied **after** jitter — a ceiling jitter can exceed is not a ceiling (this
   was a real bug caught by test at 8,007ms against an 8,000ms cap).
4. Only transient failures retry. A proof the prover rejected on its merits
   fails identically every time; retrying it burns the budget and delays the
   real error.
5. Classification must walk the error's **source chain**, not just the top
   message. `TransactionProverError` carries its cause via thiserror's
   `#[source]`, so `to_string()` yields only `"failed to prove transaction"`
   while `"Timeout expired"` sits further down. Reading the top message alone
   classified every failure as permanent and the retry silently never fired —
   caught only because failed-operation durations were unchanged
   (21,006ms vs 20,920ms p50). Keep the regression test that asserts the top
   message does *not* carry the signal, so the test cannot pass for the wrong
   reason.
6. Document the option in `docs/MULTISIG_SDK.md`.

zeljkoX:
## Summary

Closes #360.

Adds optional prover configuration to both Miden multisig SDKs, supporting:

- Configurable retry attempts for remote proving
- Custom remote prover URLs
- Consistent retry classification and backoff behavior across Rust and TypeScript
- Existing behavior when prover configuration is omitted

No Miden RPC retry behavior or GUARDIAN wire contracts are changed.

### TypeScript

```ts
const client = new MultisigClient(midenClient, {
  guardianEndpoint,
  midenRpcEndpoint,
  prover: {
    url: 'https://prover.example',
    retry: {
      maxAttempts: 4,
    },
  },
});
```

The `prover` property is optional. When omitted, the SDK inherits the prover configured on the injected `MidenClient`.

### Rust

```rust
let prover = ProverConfig::new()
    .with_url("https://prover.example")?
    .with_retry_policy(ProverRetryPolicy::new(4));

let client = MultisigClient::builder()
    .miden_endpoint(Endpoint::devnet())
    .guardian_endpoint(guardian_endpoint)
    .prover_config(prover)
    // ...
    .build()
    .await?;
```

Calling `.prover_config(...)` is optional. Existing network defaults remain unchanged.

## Retry behavior

- `maxAttempts` / `max_attempts` represents total proof attempts, not additional retries.
- Remote proving defaults to two total attempts.
- Local proving always performs one attempt and is never retried.
- Only transient proving failures are retried.
- Permanent structured errors take precedence over transient evidence anywhere in the error chain.
- Retryable HTTP statuses include `408`, `429`, `502`, `503`, and `504`.
- Retryable gRPC conditions include cancellation, deadline exceeded, unavailable, and resource exhaustion.
- Connection timeout, reset, and broken-pipe equivalents are recognized from flattened errors.
- Backoff uses exponential delay with jitter, starting at 500 ms and capped at 8 seconds.
- A custom prover failure never falls back to the default or local prover.
- The final upstream error is returned unchanged when attempts are exhausted.

## Transaction safety

Retries are scoped exclusively to proving.

TypeScript uses an explicit staged workflow:

```text
execute once → prove with retry → submit once → apply once
```

Rust installs the retry policy through the `TransactionProver` boundary, so transaction execution and submission remain outside the retry loop.

## Additional changes

- Preserves Rust prover configuration when recreating the underlying Miden client.
- Propagates resolved TypeScript prover configuration to created and loaded multisigs.
- Adds optional custom prover configuration to the Rust demo and browser examples.
- Documents defaults, retry classification, custom URL behavior, and combined configuration.
- Adds shared contract fixtures to keep Rust and TypeScript behavior aligned.

## Summary by CodeRabbit

* **New Features**
  * Added configurable transaction prover settings, including custom remote endpoints and retry budgets.
  * Remote proving now retries transient failures with bounded backoff while avoiding retries for permanent errors.
  * Added prover configuration support across Rust, TypeScript, browser, web, and demo clients.
  * Added validation for absolute HTTP(S) prover URLs and clearer configuration errors.

* **Documentation**
  * Updated SDK, README, and demo guidance with prover endpoint and retry configuration examples.


haseebrabbani:
@zeljkoX i think `speckit/features/001-prover-retry-policy/contracts/prover-policy-fixtures.json` was not committed — the Rust `include_str!` and the three TS fixture tests reference it, which is why Clippy and Test Suite are red

haseebrabbani (approved):

Issue #317, opened 2026-07-10
Define and reach Guardian scalability target: 100k guarded accounts, 100 concurrent transacting users, 20k concurrent state readers


Dominik1999:
**Parent:** Deployment & Scaling (#256)

## Goal

Establish an explicit scalability target for the Guardian server and close the gap to it:

| Dimension | Target |
|---|---|
| Guarded accounts | **100,000** |
| Concurrent users transacting (push/canonicalize deltas) | **up to 100** |
| Concurrent users reading state (`get_state`) | **up to 20,000** |

Acceptance criteria at target load:

- `get_state` p95 ≤ 1s at 20k concurrent readers.
- Delta canonicalization p95 ≤ 30s at 100 concurrent writers — acceptable only under very high load; further optimization can come later.
- Zero auth-window (replay-protection) failures.

## Where we stand (benchmark report, `crates/server/bench/reports/2026-02-15.md`)

The February 2026 benchmark runs (local machine, mostly default Postgres settings) show this target is **not yet reachable**:

- **Reads are the main gap.** At 10k concurrent readers (half the target), `state-read` completed with 0 failures but p50 was ~4.4 min and p95 ~8.6 min. At the 20k target this would degrade further.
- **Mixed load breaks the auth window.** At 10k users in `state-sync` (4 reads : 1 write), queueing pushed 14.1% of requests (7,053/50,000) past the 5-minute replay-protection window: `Authentication failed: Request timestamp outside allowed window (max 300000ms)`. These are operational failures under load, not correctness bugs.
- **Writes at target concurrency are close but slow.** The canonicalization run with 100 concurrent workers completed 1,198/1,200 ops (the 2 failures were auth-window expiries), but with p95 ~14s / p99 ~17s.
- **Account scale is untested.** Benchmarks seeded at most 10k accounts; behavior at 100k guarded accounts is unmeasured (seeding alone took ~20 min at 10k).
- At ~1000 concurrent users, runs were stable with zero failures but p95 latency was already ~15s.

## Known bottlenecks (verified still present on `main`)

1. **DB pool limit** — storage and metadata pools default to 16 connections (`GUARDIAN_DB_POOL_MAX_SIZE` / `GUARDIAN_METADATA_DB_POOL_MAX_SIZE`; prod Terraform sets 32). Far too low for a 20k-reader fan-out.
[...]

PR #368, opened 2026-07-31
#365 read path auth write perf improvements

### LINKED ISSUE #365: Read path: every authenticated read performs a durable auth write
## Summary

Every authenticated read issues a durable `UPDATE` to `account_metadata` for
replay protection. That write accounts for **33.5% of read-path database time**
(authentication overall: 43.5%), and removing it measured **+71% throughput**.
Postgres saturates while both Guardian replicas idle, so the read path cannot be
scaled by adding server replicas — the fix has to make the database cheaper per
read.

**This is not specific to `get_state`.** The write lives in the shared
authentication path, so *all five* signature-authenticated read endpoints pay it
identically — including the two that SDK clients poll in a loop. One fix, one
call site, five endpoints improved.

The replay-protection guarantee is correct and must be kept. Only the *storage
placement* of the hot column needs revisiting.

## Every authenticated read pays this, not just `get_state`

The cost is in `resolve_account` (`crates/server/src/services/mod.rs:101-190`),
which every account-scoped endpoint calls. Reads that pay the metadata `SELECT`
plus the CAS `UPDATE`:

| Endpoint | Service | Client usage pattern |
| --- | --- | --- |
| `get_state` | `services/get_state.rs:25` | the #317 target metric |
| `get_delta_since` | `services/get_delta_since.rs:33` | **SDK sync loop — polled** |
| `get_delta_proposals` | `services/get_delta_proposals.rs:28` | **cosigner queue — polled** |
| `get_delta_proposal` | `services/get_delta_proposal.rs:29` | per-proposal fetch |
| `get_delta` | `services/get_delta.rs:27` | status / abandon polling |

The write endpoints (`push_delta`, `push_delta_proposal`, `sign_delta_proposal`,
`abandon_candidate`) go through the same path, but there the write is
unremarkable — those transactions were writing anyway.

`get_state` is the most important endpoint *by benchmark definition*, but not
necessarily by production call volume: `get_delta_since` is how every SDK client
syncs and `get_delta_proposals` is how cosigners discover work, so both are
polled repeatedly. On a deployment with active multisig accounts they can rival
or exceed `get_state` in call count. Any measurement that looks only at
`get_state` understates the aggregate cost of this write.

**Paths that do not pay it**, for contrast:

- `/state/lookup` — deliberately excluded. `services/lookup_account.rs:16-19`
  documents why: there is no `account_id` to key per-account replay state
  against, so it enforces only the timestamp skew window. Worth noting as
  precedent — the system already accepts a bounded replay window on one
  authenticated path.
- `/pubkey`, `/status`, `/` — unauthenticated.
- Dashboard reads — operator session cookie via `authenticate_session`
  (`dashboard/middleware.rs:18-21`), which performs no write. Those endpoints
  have a separate cost problem (per-row summary computation) that does not belong
  in this issue.

## Evidence

Measured 2026-07-30 against `0.16.0` / `5d3999eadbf0`: two Guardian replicas
behind a round-robin proxy, shared Postgres (1.5 CPU cap), pools 16/16.
Workload: closed-loop `get_state`, 100% ECDSA.

`pg_stat_statements`, 128-reader leg (100s, 231,469 reads recorded):

| Query | Calls | Total ms | Share of DB time |
| --- | --- | --- | --- |
| `SELECT states` (the actual read) | 280,100 | 59,687 | 56.5% |
| **`UPDATE account_metadata SET last_auth_timestamp`** | 280,100 | 35,368 | **33.5%** |
| `SELECT account_metadata` | 280,264 | 10,583 | 10.0% |

Authentication is therefore 43.5% of database time, and the write alone costs
more than half of what the read it authenticates costs.

Throughput and latency across concurrency levels — flat-to-declining as
concurrency rises, i.e. a saturated system with a ceiling around 2,300/s:

| Readers | Throughput | p50 | p95 | p99 |
| --- | --- | --- | --- | --- |
| 768 | 2,090/s | 371ms | 760ms | 1,429ms |
| 256 | 2,151/s | 109ms | 221ms | 344ms |
| 128 | **2,314/s** | 52ms | **101ms** | 165ms |

**Postgres is the constraint, not Guardian.** Postgres peaked at ~165% against
its 150% CPU cap in every leg, while the two Guardian replicas stayed well below
their limits. Adding server replicas cannot move this ceiling.

### Causal check: the write patched out

To establish causation rather than correlation, the CAS was patched out, the
server rebuilt (tagged `NOCAS-EXPERIMENT`), and the identical profile re-run. The
patch was reverted immediately — it disables replay protection and exists only as
a measurement device.

| | With CAS | CAS removed | Change |
| --- | --- | --- | --- |
| Throughput | 2,314/s | **3,957/s** | **+71%** |
| p95 | 101ms | 66ms | −35% |
| DB time per read | 0.385ms | **0.214ms** | −44% |
| Postgres peak CPU | 165.7% | 158.9% | ~unchanged |

The 44% drop in per-read DB cost exceeds the 33.5% the write consumes directly,
because removing it also removes the MVCC churn it caused: the state `SELECT`
sped up from 0.213ms to 0.178ms mean on the same rows.

**Note that Postgres stayed pegged.** Removing the write does not move the
bottleneck elsewhere — it lets the same bottleneck serve 71% more reads.

## Mechanism

The write is the replay-protection compare-and-set
(`crates/server/src/metadata/postgres.rs:267-299`, called from
`crates/server/src/services/mod.rs:166`):

```sql
UPDATE account_metadata
   SET last_auth_timestamp = $1, updated_at = $2
 WHERE account_id = $3
   AND (last_auth_timestamp IS NULL OR last_auth_timestamp < $1)
```

A signature proves a request is authentic, not fresh. The timestamp skew window
bounds how long a captured request stays usable; the CAS makes each timestamp
usable exactly once. With multiple replicas behind a proxy, a replica-local cache
could be defeated by replaying to another replica, so the state must be durable,
atomic, and shared — a database write. **The guarantee is not in question here.**

The cost comes from where the column lives
(`crates/server/migrations/2026-03-12-000002_account_metadata/up.sql`):

```sql
CREATE TABLE IF NOT EXISTS account_metadata (
    account_id VARCHAR(128) PRIMARY KEY,
    auth JSONB NOT NULL,              -- signer public keys
    network_config JSONB NOT NULL,
    created_at TIMESTAMPTZ NOT NULL,
    updated_at TIMESTAMPTZ NOT NULL,
    has_pending_candidate BOOLEAN NOT NULL DEFAULT FALSE,
    last_auth_timestamp BIGINT        -- the hot column
);

CREATE INDEX IF NOT EXISTS idx_metadata_pending
    ON account_metadata(has_pending_candidate)
    WHERE has_pending_candidate = TRUE;
```

Postgres is MVCC: an `UPDATE` writes a whole new row version. Changing eight
bytes rewrites the entire tuple, including `auth` — which for a Falcon multisig
holds Falcon-512 public keys at roughly 900 bytes each. Hence WAL volume driven
by the full row width, continuous dead tuples on a nominally read-only workload,
and HOT updates not guaranteed (wide rows, plus the partial index on
`has_pending_candidate`).

Two secondary effects: the CAS also bumps `updated_at`, so that column tracks
"last authenticated read" rather than "last configuration change"; and autovacuum
runs continuously against a workload an operator would reasonably expect to
generate no garbage.

## History

- `ae78e3b` (2026-01-22, #104, "fix: prevent replay attacks using timestamp")
  introduced the CAS. A security fix, and the guarantee is sound.
- The column was placed on the existing `account_metadata` table — the natural
  choice at the time, with a cost invisible at schema level and visible only
  under MVCC at high read rates.
- `5e32f60` (#124, ECDSA support) widened the `auth` JSONB that every read now
  rewrites.

Nothing here was a mistake when made; the price became visible only once the read
path was measured.

## Impact on the #317 target and the production shape

The #317 target is `get_state` p95 ≤ 1s at 20,000 concurrent readers, but that is
a **paced** definition — one `get_state` per user per 10s — so offered load is
**2,000 reads/s, not 20,000**. At ~100ms latency that is roughly 200 requests in
flight. (Stating this explicitly because reading the target as 20,000 reads/s
makes the gap look like an order of magnitude when it is not.)

Against production as committed in Terraform (`db.r6g.large`, 2 vCPU / 16 GiB,
RDS Proxy):

- **Throughput: plausible but thin.** 2,000/s against a measured ~2,300/s on
  1.5 CPU is ~86% utilisation — on a read-only workload with nothing else
  running. Production shares the instance with pushes, canonicalization workers,
  dashboard queries, and the four polled read endpoints above. Cheapening the
  auth write moves this to roughly 50%.
- **Connections: a non-issue.** ~200 in-flight requests, and RDS Proxy absorbs
  idle client connections. The committed prod pool default
  (`guardian_db_pool_max_size = 32`, `infra/data.tf:151`) is comfortably sized.
- **Memory: fine.** 100k accounts of states plus metadata fits well inside
  16 GiB.
- **Storage IOPS: unverified and worth checking.** `infra/rds.tf:20-41` sets no
  `storage_type`, so the instance takes the provider default (gp2), and the
  committed prod `allocated_storage` is **50 GB** (`infra/data.tf:129`). On gp2
  that is 150 baseline IOPS, burstable to 3,000 on depletable credits. Every read
  issuing a durable write makes this path IOPS-sensitive in a way a normal read
  path is not. WAL commits group-fsync, so it is not a hard 1:1 with
  transactions — but with dirty-page flushing and autovacuum churn from the CAS,
  gp2 baseline is uncomfortably close. gp3 gives 3,000 IOPS baseline and is
  provisionable.

> **Confirm two deployed values.** An earlier draft of this analysis assumed
> `rds_allocated_storage = 100` and `guardian_db_pool_max_size = 64`. The
> committed defaults are 50 and 32. If the live instance really is 100 GB with a
> pool of 64, those are tfvars or console overrides rather than repo state —
> worth confirming, since the gp2 IOPS figure halves at 50 GB.

## Options

| Option | Effect | Cost / caveat |
| --- | --- | --- |
| **Split the hot column into its own narrow table** — `account_auth_state(account_id PK, last_auth_timestamp)` | Row width drops from 1KB+ to ~24 bytes; WAL volume collapses; HOT updates become possible | No change to the security property — the CAS is unchanged. Needs a migration plus backfill. Recovers a fraction of the +71%, not all of it. |
| **Move the auth state off the SQL primary** (e.g. ElastiCache / DynamoDB) | Removes the write from the database entirely — approaches the measured +71% upper bound | New runtime dependency, and the auth state may be lost within the skew window after a crash, reopening a bounded replay window. A security decision, not an optimisation. |
| **Coarsen granularity** — persist only when advancing by more than N ms | Fewer writes | Replays within N ms become possible. Weakens the guarantee by construction, though `/state/lookup` shows the project already accepts a bounded window on one path. |

The first is the low-risk win and should be measured before the others are
argued about: it is a migration with no change to the security property, and it
tests the row-width hypothesis directly.

## Out of scope

- **Scaling reads with RDS read replicas.** Not addressed here on purpose — a
  read that writes cannot run on a read-only replica at all, so replica work is
  blocked on this issue regardless, and it carries its own infrastructure
  questions. It deserves a separate issue once the write is cheap.
- **Dashboard read cost** (per-row summary computation) — different mechanism,
  different path, no CAS involved.
- **gp2 → gp3 storage migration** — stands alone from this work; worth its own
  infra issue.

## Reproduction

```bash
cd benchmarks/diagnostic-stack
docker compose --env-file .env --env-file variants/read-budgeted.env up -d
TOKIO_WORKER_THREADS=4 ./run-diag.sh <label> profiles/diag-read-128.toml
```

Baselines:

- `results/read-128-t4-20260730T083554Z` — with CAS, 2,314/s
- `results/nocas-128-t4-20260730T092323Z` — CAS removed, 3,957/s (the upper bound
  any fix can approach)

## Caveats

- The share-of-time figures and the +71% are **measured**. That **row width is
  the mechanism** is well-supported but not yet benchmarked — splitting the table
  is the experiment that confirms or refutes it.
- **+71% is an upper bound.** Splitting the table keeps the write and only makes
  it cheaper.
- Every leg was flagged `OVERSUBSCRIBED` (peak host load 12.4–17.4 on 10 cores),
  so the latency figures are one-sided bounds, not verdicts. The A/B comparison
  is still sound — both legs ran under identical contention.
- All legs drove `get_state` only. The other four authenticated reads share the
  same authentication path and therefore the same per-call write, but their
  *total* cost was not measured — their own query shapes differ.
- Numbers came from Postgres in a VM on Apple Silicon; storage and CPU behaviour
  differ from Graviton2. Treat them as order-of-magnitude checks.
- Full context:
  `speckit/features/317-scalability-target-measurement/test-report-2026-07-30.md`
  (§7.4, §7.5, §7.7).

## Suggested next steps

- [ ] Split `last_auth_timestamp` into a narrow table and re-run the 128-reader
      leg — the cheap experiment that converts the mechanism from well-supported
      to measured
- [ ] Re-measure with a mixed read profile (`get_delta_since` /
      `get_delta_proposals` alongside `get_state`) so the aggregate saving is
      known rather than inferred from one endpoint
- [ ] Confirm the live RDS `storage_type` and `allocated_storage`
- [ ] Decide whether the residual cost after the split justifies moving auth
      state off the SQL primary — a security review, not just a perf review

zeljkoX:
Closes #365.

Authenticated reads were writing replay-protection state (`last_auth_timestamp`)
into the wide `account_metadata` row — ~33% of read-path DB time in the #317
diagnostic runs, plus constant churn/`updated_at` coupling on a row that config
changes and dashboards also touch.

This splits replay state into a dedicated narrow table, `account_auth_state`
(`account_id` PK, `last_auth_timestamp`, fillfactor 50), updated via a raw
`INSERT … ON CONFLICT DO UPDATE … WHERE <` CAS upsert. The filesystem backend
gets an equivalent `auth_state.json`. The replay guarantee is unchanged: each
per-account timestamp is accepted exactly once, durably, fleet-wide.

Migration backfills under an `ACCESS EXCLUSIVE` lock and drops the old column;
replicas still on the old binary fail closed during a rolling deploy (no replay
window). A missing `auth_state.json` on a migrated filesystem store also fails
closed.

**Benchmark A/B** (diagnostic stack, 128 readers, `get_state`, same machine,
fresh DB per leg — measured on the #317 integration branch, where the harness lives):

| | main | this PR |
|---|---|---|
| Throughput | 2,836/s | 3,361/s (+18.5%) |
| p95 | 90.4ms | 80.7ms |
| Auth DB time per read | 120.6µs | 64.8µs (**−46%**) |
| `account_metadata` updates during leg | 338k | **0** |
| Replay writes HOT | — | 100% (400k/400k) |

Residual: the upsert itself is ~23% of read DB time — the cost of a durable
write per authenticated read. Closing that requires weakening the guarantee
(coarsening/UNLOGGED/external store), which this PR keeps out of scope.

## Summary by CodeRabbit

* **New Features**
  * Replay-protection timestamps are now stored separately from account metadata.
  * Authentication replay checks now reject repeated or older requests consistently across services.
  * Concurrent timestamp updates are handled safely, including persistence across metadata changes.

* **Bug Fixes**
  * Prevented replay protection from being silently reset when authentication state is missing or migrations are interrupted.

* **Documentation**
  * Added guidance for backups, recovery, startup failures, and rolling deployments involving authentication state.


zeljkoX:
@haseebrabbani Just a reminder to review once you find time. Thanks!

haseebrabbani (approved):
Looks good, minor comments:

1. **`down.sql` lacks the lock discipline `up.sql` has.** The backfill `UPDATE` → `DROP TABLE account_auth_state` sequence never locks `account_auth_state` — the `ALTER TABLE` only locks `account_metadata`, which doesn't block new-binary CAS writes. Timestamps committed between the backfill snapshot and the `DROP TABLE` are silently lost — the mirror image of the hazard `up.sql`'s comment explains. Suggest `SET LOCAL lock_timeout` + `LOCK TABLE account_auth_state IN ACCESS EXCLUSIVE MODE` before the backfill.
2. **The new Postgres tests never run in CI** (`cargo test --lib`, no `DATABASE_URL`, no `--ignored` pass). Matches existing practice, but the migration-backfill test is the one guarding against replay-state loss — can you confirm it was run locally? A Postgres service container in CI could be a follow-up.
3. **Filesystem CAS existence check is advisory** — it takes the cache read lock, drops it, then takes the auth-state write lock. Harmless today (no delete op exists; `resolve_account` checks existence first), but a comment noting the race would prevent surprise if deletion is ever added.

zeljkoX:
> Looks good, minor comments:
> 
> 1. **`down.sql` lacks the lock discipline `up.sql` has.** The backfill `UPDATE` → `DROP TABLE account_auth_state` sequence never locks `account_auth_state` — the `ALTER TABLE` only locks `account_metadata`, which doesn't block new-binary CAS writes. Timestamps committed between the backfill snapshot and the `DROP TABLE` are silently lost — the mirror image of the hazard `up.sql`'s comment explains. Suggest `SET LOCAL lock_timeout` + `LOCK TABLE account_auth_state IN ACCESS EXCLUSIVE MODE` before the backfill.
> 2. **The new Postgres tests never run in CI** (`cargo test --lib`, no `DATABASE_URL`, no `--ignored` pass). Matches existing practice, but the migration-backfill test is the one guarding against replay-state loss — can you confirm it was run locally? A Postgres service container in CI could be a follow-up.
> 3. **Filesystem CAS existence check is advisory** — it takes the cache read lock, drops it, then takes the auth-state write lock. Harmless today (no delete op exists; `resolve_account` checks existence first), but a comment noting the race would prevent surprise if deletion is ever added.

Thanks for feedback! Applied! Tests are passing locally! Definitely we should improve tests pipelines to run all. Will create issue for this!

Issue #351, opened 2026-07-23
Archive/unarchive for pending proposals (reversible retraction) — phase 2 of #337


haseebrabbani:
## Background

#337 was split in two after the discussion there:

- **Phase 1** (PR #350): stale proposals — those built on a superseded `prev_commitment` — no longer count toward `GUARDIAN_MAX_PENDING_PROPOSALS_PER_ACCOUNT`, fixing the permanent `PendingProposalsLimit` lockout.
- **Phase 2** (this issue): a way for a proposer to *deliberately* retract a live, still-viable proposal — e.g. withdrawing it before cosigners sign — via a reversible archive/unarchive mechanism rather than hard deletion.

Design sketch below is from @zeljkoX's comment on #337.

## Proposed design

- Replace deletion with reversible **archive/unarchive**.
- Store `archived_at` separately from `DeltaStatus` — archival is proposal metadata, not a delta lifecycle state.
- Only the original proposer may archive or unarchive.
- Archived proposals cannot be signed or submitted and do not consume the pending-proposal limit.
- Unarchive succeeds only if the proposal remains viable (derived from current account state — primarily `prev_commitment` plus network validation), no candidate exists, and capacity is available.
- Archive / sign / unarchive / candidate creation must use atomic conditional storage operations to prevent races.
- Support both Postgres and the default filesystem backend, plus HTTP and gRPC, so the Rust and TS clients stay aligned.
- Hard deletion is optional retention work on top of this — e.g. purge archived payloads after 30 days while retaining a minimal audit tombstone.

## Open questions

- **Proposer identity/authz:** nothing today ties a proposal to an authenticated proposer identity (`DeltaStatus::Pending` carries a `proposer_id`, but it isn't verified against the caller's credentials). "Only the original proposer may archive" needs an authz story before implementation.
- Whether archived proposals should appear in `pull_pending_proposals` (flagged) or only via a dedicated listing.
- Client API shape (`multisig.archiveProposal(commitment)` / `unarchiveProposal(commitment)`?) and how it relates to the abandon flow from #327.

PR #370, opened 2026-08-03
RFC: 254 Guardian prove and commit transaction

zeljkoX:
## Summary

RFC proposing server-side Miden transaction execution in Guardian.

## Main decision

Guardian assembles the transaction witness, delegates proving to a remote
prover, submits the proven transaction, and observes the final outcome.

## Review focus

- Overall architecture and security assumptions
- Chain-view and authenticated-note assembly
- Execution lifecycle and no-retry boundary
- Questions intended for the Miden team
- Missing cases or unclear sections

## Scope

This PR contains the RFC and supporting research/validation artifacts.
It does not implement the complete feature.

## RFC

[Read the rendered RFC](https://github.com/OpenZeppelin/guardian/blob/254-guardian-prove-and-commit/docs/rfcs/0001-server-side-transaction-execution.md)

Please leave review feedback and proposed changes in this PR.

## Summary by CodeRabbit

* **New Features**
  * Added support for assembling authenticated blockchain and note data for transaction execution.
  * Added RPC capabilities for chain synchronization and paginated note synchronization.
  * Added execution data handling for account, transaction, and blockchain inputs.
  * Introduced configurable Guardian-managed proving and execution modes.

* **Documentation**
  * Added comprehensive API contracts, SDK guidance, quickstart instructions, RFC documentation, and implementation plans for Guardian prove-and-commit execution.

* **Tests**
  * Added extensive integration and live-network validation for synchronization, transaction execution, proving, signing, and expiration behavior.


docs/rfcs/0001-server-side-transaction-execution.md
```diff
@@ -0,0 +1,489 @@
...
+  Guardian->>Node: Submit transaction to Miden Node
+
+  Note over Thin,Guardian: 4. Outcome Polling
+  loop Poll Status
+    Thin->>Guardian: GET /delta/proposal/execution
+    Guardian-->>Thin: state: proving → submitted → landed
+  end
+```
+
+### 1.2 End-to-End Execution Lifecycle & State Machine
```
bobbinth, line 130: Thank you for putting this together. I haven't gone through the full proposal yet, but want to make sure that I understand the overall approach first. One area I'm specifically interested in is how we'd handle conflicting requests and potential attack vectors associated with them.

My current understanding is as follows:

1. A user can submit a transaction request to the Guardian.
2. The guardian will verify that this request is valid - this includes making sure that the request builds on the latest known state of the account, that it is signed by one of the signers etc.
3. Once the validity is verified, the Guardian will convert the request into a transaction to derive the new account state.
4. Then, one of two things can happen:
  a. If there not enough signatures for the request, the Guardian will wait until sufficient number of signatures is collected. I'm assuming there will be a dedicated endpoint for this. Also, I'm assuming clients will be able to fetch the latest account state + the associated transaction data to make sure they sign it (or do they fetch transaction requests and then execute them locally?)
  b. Once the guardian has enough signatures, it will sign the transaction.
5. Once the transaction is signed, the Guardian will execute it, and submit it to the network.
6. Once the network commits the transaction, the Guardian will know that the transaction is committed and will mark it accordingly locally.

Assuming the process is like so (and please correct if not), I have the following questions:

1. What happens if while the Guardian is in state 4a (i.e., the transaction request is accepted, but does not yet have enough signatures), another, potentially conflicting, transaction request is submitted? Would the pending request be replaced? Would the incoming request be rejected? Or would there be some kind of logic to resolve this? I'm asking this because depending on what we do here, some signers may try to block the account by submitting requests that other signers may not want to sign.
2. How would the Guardian handle transaction request on top of currently executing transaction requests. That is, let's say request 1 takes account from state `A` to `B`, but before the Guardian has had a chance to prove this transaction, another request that takes the account from `B` to `C` arrives. Would the Guardian reject this request? Or build a set of chained requests? The latter is more complicated but also leads to a much better UX on the wallet side (something we'd want to enable).
3. What are the states in which account on the Guardian could be in? I'm assuming there is a `committed` state (i.e., the state on the Guardian is the same as onchain) - but what other states would we have? This is especially interesting in the context of multiple in-flight requests for the same account. For example, using the above scenario, state `A` may be `committed`, state `B` may be `proving`, and state `C` could be `validated` - or something like that.

zeljkoX: Thanks, you're close, but proposal and execution are two separate phases. Guardian can store multiple proposals and collect signatures without reserving the account. It only creates a reservation after a proposal has reached its effective threshold and a cosigner explicitly asks for delegated execution.

**Flow:**

* A Miden-capable party builds the `TransactionRequest`, runs it at an authenticated reference-block anchor to get the `TransactionSummary`, and creates a proposal that carries the request, the summary, and the chain anchor.
* Guardian authenticates the proposer as a registered cosigner, validates the summary against the current canonical state, checks the chain-anchor binding, tags the proposal with its base commitment, and stores it. It does not reserve the account or start delegated execution yet.
* Cosigners fetch the proposal and sign the summary commitment. A thin cosigner can sign mechanically with just its key, but it cannot independently verify that the metadata it sees matches the transaction it is signing. A Miden-capable cosigner can rebuild the transaction locally and verify the commitment before signing. That is what the current multisig SDKs do.
* Once the effective threshold is met, any registered cosigner can explicitly request delegated execution.
* Guardian then creates the per-account execution reservation, reproduces the transaction using the proposal's authenticated reference-block anchor, confirms the canonical state still matches the proposal's base state, verifies the result matches the signed summary, and validates the collected signatures.
* Guardian adds its own acknowledgment, executes the authorized transaction, proves it via the remote prover, atomically admits the candidate delta with the submission evidence, and submits the proven transaction.
* The candidate becomes canonical once Guardian observes it has landed on chain.

**1) Multiple proposals can coexist**

A new proposal does not replace or reject an existing pending proposal, and creating a proposal does not lock the account. Only a proposal that has reached threshold and enters delegated execution gets the per-account reservation.

If two proposals become executable, the first execution request to acquire the reservation goes ahead. If a different proposal tries to execute while that reservation is active, it is refused synchronously with `GUARDIAN_EXECUTION_CONFLICT` and the response tells you which proposal is blocking. Re-triggering the same proposal is idempotent and just returns the existing execution. Direct self-execution through `POST /delta` is also refused while the reservation is active, so delegated and client-side execution cannot race.

A competing proposal based on the same canonical state can still be created while the first one is proving and before its candidate is admitted, but it cannot enter execution while the reservation is held. Once a candidate exists, creation of further proposals is blocked by the existing pending-candidate conflict. When one proposal lands, any proposals built on the previous state become stale, cannot execute, and no longer count against the viable pending-proposal limit.

Proposal creation is restricted to authenticated multisig signers, which prevents public spam. Guardian also caps the number of viable pending proposals per account. The abandon endpoint handles stuck candidate deltas after admission. It does not remove below-threshold pending proposals. Issue #351 proposes an archive/unarchive flow that would let the original proposer retract a pending proposal without hard deleting it, and archived proposals would not count toward the limit.

Together this prevents public spam, bounds storage, and gives a recovery path for accidental or obsolete proposals. It does not fully eliminate denial of service by a malicious or compromised authorized signer. Such a signer could still fill the proposal allowance and refuse to archive. That remaining insider liveness risk could be closed with a per-proposer quota, automatic expiry, or threshold-authorized archival.

**2) No chained proposals in v1**

In v1 we do not support chained proposals like `A -> B -> C`. Until `A -> B` lands, Guardian's canonical state stays at `A`. `B` is only the expected result of the in-flight transaction.

Guardian validates every new proposal against its current canonical state. So a real `B -> C` proposal submitted while canonical state is still `A` is rejected at creation time. Competing proposals based on `A` can coexist, but only one can enter execution.

A proposal that was valid at creation can still become stale before execution, for example because another proposal advanced the account first. Guardian therefore rechecks state and bindings when delegated execution starts. If the account no longer matches the proposal's base state, execution fails before proving or submission.

Chaining is a deliberate v1 non-goal, not a fundamental limitation. The current model is built around one canonical account state, at most one active candidate per account, a no-retry boundary for submission, and resolution by observing the chain. Supporting chains would require explicit dependencies, speculative intermediate states, ordering rules, and cascading invalidation when a parent is rejected, superseded, or expires.

Miden 0.16 also binds the reference block commitment and expiration delta into the signed summary. That makes advance signing of `B -> C` harder, because the child transaction cannot just be rebuilt later against a newer reference block without changing what cosigners signed. Chained wallet flows would likely need a separate intent or batching layer rather than just queueing today's transaction summaries.

**3) Status lives on different objects**

Guardian tracks status at three levels:

* **Proposal:** stored as `pending`. Whether it looks ready on the client is derived from collected signatures. Before execution, Guardian recomputes readiness itself using only distinct, valid signatures from currently registered cosigners and the effective threshold.
* **Execution:** `pending`, `proving`, `submitted`, `landed`, or `failed`.
* **Delta:** `candidate`, `canonical`, `retained`, or `discarded`.

The account itself always has exactly one canonical state.

Taking `A -> B` as an example:

* While signatures are being collected, the account remains canonical at `A`. The proposal is pending and may appear ready once enough signatures are present.
* Once delegated execution is accepted, the account is still canonical at `A`, while execution moves through `pending` and `proving`. No candidate delta exists yet and `B` is only the expected result.
* When Guardian crosses the no-retry boundary, execution becomes `submitted` and an `A -> B` candidate delta exists. The canonical state is still `A`.
* If Guardian observes the expected commitment on chain, the candidate becomes canonical, the account advances to `B`, and execution becomes `landed`.
* If the node definitively rejects it, the account moves to a different commitment, or the transaction expires while the account stays at `A`, execution becomes `failed`. The candidate is discarded or removed and the reservation is released.
* A `retained` delta is an unresolved candidate that has been removed from the active blocking path but kept around for reconciliation.

So we would not describe `B` as an account state that is "proving", or `C` as "validated". Those statuses belong to proposals, executions, and deltas, not to speculative account states. In v1 the model is one canonical account state plus at most one active candidate transition, because chained in-flight transactions are not supported.

bobbinth: Thank you very much for the reply - everything makes sense. A few more questions/comments:

> Guardian authenticates the proposer as a registered cosigner, validates the summary against the current canonical state, checks the chain-anchor binding, tags the proposal with its base commitment, and stores it. It does not reserve the account or start delegated execution yet.

What does "validates the summary against the current canonical state" mean? Does the guardian build a transaction based on the specified `TransactionRequest` and then compare the summary to the one provided? If so, why not just get the transaction request from the user and build the summary (and they check the provided signature against this summary)?

> Once the effective threshold is met, any registered cosigner can explicitly request delegated execution.

Our typical wallet set up is 2-of-3 multisig with the guardian being one of the signers. So, once a transaction request from such wallet arrives, it will be immediately executable by the guardian. But I guess the wallet would still need to perform this 2-step process - i.e., submit the request (with all required signatures), request execution from the guardian - right?

> Proposal creation is restricted to authenticated multisig signers, which prevents public spam. Guardian also caps the number of viable pending proposals per account.

We might want to enforce the cap on per account and signer combination (e.g., the same signer cannot submit more than 2 proposals for a given account). This would prevent a malicious signer from saturating the proposal pool so that nobody else can submit a proposal.

> Chaining is a deliberate v1 non-goal, not a fundamental limitation. The current model is built around one canonical account state, at most one active candidate per account, a no-retry boundary for submission, and resolution by observing the chain. Supporting chains would require explicit dependencies, speculative intermediate states, ordering rules, and cascading invalidation when a parent is rejected, superseded, or expires.

This makes sense and is probably the right call for v1. In the longer term, we will need to support this imposes undesirable UX constraints in a couple of ways:

- Transaction latency: the wallet must wait until the guardian can observe the committed transaction on-chain before sending the next transaction to the guardian. This could potentially take 5+ second.
- Strict ordering: in larger multisigs it may be desirable to be able to execute transactions "out-of-order". For example, there could be two separate proposals: "send 10 coins to X" and "send 5 coins to Y". These are not conflicting proposals and whether one gets executed before the other doesn't really matter. So, if the "send 10 coins" gets all required signatures collected first, executing/proving/committing it shouldn't invalidate the "send 5 coins" proposal. Doing this should be possible because transaction summary does is not bound to a specific account state - it contains only the transaction delta which describes an incremental change to the account.
- Limited batching: in the future, the guardian would be able to batch multiple transactions together. In the above example, both "send 5 coins" and "send 10 coins" transactions could end up in a single batch which will result in a single update on the network level.

As mentioned above, this is something to do beyond v1 - but we should make sure to keep these considerations in mind when coming up with future designs.

> **Execution:** `pending`, `proving`, `submitted`, `landed`, or `failed`.

I would consider renaming `landed` into `committed` as that's the terminology we use elsewhere.

zeljkoX: > What does "validates the summary against the current canonical state" mean? Does the guardian build a transaction based on the specified TransactionRequest and then compare the summary to the one provided? If so, why not just get the transaction request from the user and build the summary (and they check the provided signature against this summary)?

Guardian does not execute the `TransactionRequest` when accepting a proposal. The current check parses the supplied summary and verifies that the stored account state matches its recorded commitment. Guardian records that commitment as the proposal’s base. Reproducing the request and comparing the resulting summary with what cosigners signed happens when delegated execution starts.

For v1, I’d keep that flow: the client builds the transaction request, derives the summary, and submits both with its signature when opting into delegated execution. This RFC moves execution, proving, and submission to Guardian while preserving client-side execution as the default.

Having Guardian derive the summary makes sense as an extension. We could add an optional `prepare` endpoint that accepts the transaction request and required execution context, derives the summary, and creates an unsigned proposal. It would return the proposal ID and summary for review and signing, with signatures uploaded through the existing signing endpoint. That would let clients delegate preparation too, while keeping the current client-side proposal flow available.

> Our typical wallet set up is 2-of-3 multisig with the guardian being one of the signers. So, once a transaction request from such wallet arrives, it will be immediately executable by the guardian. But I guess the wallet would still need to perform this 2-step process - i.e., submit the request (with all required signatures), request execution from the guardian - right?

Yes. In the proposed API, the wallet submits the proposal with its signatures and then explicitly requests execution. If the proposal already meets the required threshold, those calls can happen consecutively without another signature-collection round.

> We might want to enforce the cap on per account and signer combination (e.g., the same signer cannot submit more than 2 proposals for a given account). This would prevent a malicious signer from saturating the proposal pool so that nobody else can submit a proposal.

Agreed.

> This makes sense and is probably the right call for v1. In the longer term, we will need to support this imposes undesirable UX constraints in a couple of ways:
> 
> Transaction latency: the wallet must wait until the guardian can observe the committed transaction on-chain before sending the next transaction to the guardian. This could potentially take 5+ second.
> Strict ordering: in larger multisigs it may be desirable to be able to execute transactions "out-of-order". For example, there could be two separate proposals: "send 10 coins to X" and "send 5 coins to Y". These are not conflicting proposals and whether one gets executed before the other doesn't really matter. So, if the "send 10 coins" gets all required signatures collected first, executing/proving/committing it shouldn't invalidate the "send 5 coins" proposal. Doing this should be possible because transaction summary does is not bound to a specific account state - it contains only the transaction delta which describes an incremental change to the account.
> Limited batching: in the future, the guardian would be able to batch multiple transactions together. In the above example, both "send 5 coins" and "send 10 coins" transactions could end up in a single batch which will result in a single update on the network level.
> As mentioned above, this is something to do beyond v1 - but we should make sure to keep these considerations in mind when coming up with future designs.

Agreed on all three.

> I would consider renaming landed into committed as that's the terminology we use elsewhere.
Agreed.

PR #373, opened 2026-08-04
feat: Improve RPC handling

### LINKED ISSUE #364: Miden RPC configuration and resilience: custom endpoint, configurable retry policy, retries on idempotent calls
## Summary

Three related gaps in how Guardian and its SDKs talk to a Miden **node**, all
surfaced by the #317 scalability measurement work:

1. **The server cannot be pointed at a custom RPC endpoint at all.** The URL is
   hardcoded per `NetworkType`, so a self-hosted node, a private RPC with a
   higher rate limit, or a node on any host other than `localhost:57291` is
   unreachable.
2. **Node RPC timeout and retry policy is hardcoded**, in two different places
   with two different values, and is not configurable from either the server or
   the SDKs.
3. **Idempotent RPC reads are never retried.** A single transient blip or a
   rate-limit response from the node fails the operation outright, even though
   the call is safe to repeat.

Plus a fourth item to design for now and build later: **multiple RPC endpoints
with failover** on the server (§5).

PR #362 solved the equivalent problem one layer over — the transaction **prover**
— and established the config shape, classification model, and test conventions
this work should reuse rather than reinvent (§2). #362's docs state "This policy
does not alter or retry Miden RPC requests"; that sentence is exactly the gap
this issue closes, and it must be updated as part of this work.

### Boundary with #360 / PR #362

| | #360 / PR #362 | This issue |
|---|---|---|
| Target | Transaction **prover** (a separate remote service) | Miden **node RPC** |
| Config | `ProverConfig` / `prover: {...}` | `RpcConfig` / `rpc: {...}` |
| Server | not applicable — the server does not prove | in scope: the server has no RPC config at all |
| Retried | remote proving | idempotent node reads |

They share the backoff and classification primitives (§2). #362 lands first;
this issue consumes what it produces.

### Server — no endpoint configuration

- `NetworkType::rpc_endpoint()` hardcodes three URLs:
  `crates/server/src/network/mod.rs:315-321`.
- `NetworkType` is a closed three-variant enum (`MidenTestnet`, `MidenDevnet`,
  `MidenLocal`) and `from_env` accepts only those names:
  `crates/server/src/network/mod.rs:276-281`, `306-313`.
- So `GUARDIAN_NETWORK_TYPE=MidenLocal` means *exactly* `http://localhost:57291`
  — not a node in a sibling container, not another host, not another port.
- `MidenRpcClient::connect()` is eager, with a hardcoded 30s per-request timeout
  and no retry: `crates/miden-rpc-client/src/lib.rs:15`, `23-38`. A node blip at
  boot fails server startup. (`lazy_unconnected` at `:44` shows the lazy
  alternative exists, but it is test-only.)
- **In-repo precedent:** the EVM surface already does what is being asked for
  here — per-chain RPC URLs parsed from env with URL validation
  (`crates/server/src/evm/config.rs:13-50`). The Miden surface should not be less
  configurable than the EVM one.

### Rust SDK — the custom-endpoint path is second-class

A custom node endpoint is accepted (`miden_endpoint(Endpoint)`,
`crates/miden-multisig-client/src/builder.rs`), but the custom branch is not
equivalent to the presets: the gRPC timeout is hardcoded —
`.grpc_client(endpoint, Some(20_000))` — with no way to change it, and no retry
of any kind wraps the resulting client. #362 leaves that line untouched, because
it configures the prover, not the node connection.

### TypeScript SDK

`midenRpcEndpoint` is a **required** explicit URL with no presets
(`packages/miden-multisig-client/src/raw-client.ts:28-51`). After #362 the client
options object is the natural home for an `rpc` sibling to `prover`.

So the three surfaces disagree three ways today: TS demands an explicit URL, Rust
offers presets plus a degraded custom path, and the server offers no URL
configuration at all.

### What the #317 runs measured

- The node advertises `x-ratelimit-limit: 128`. At 32 concurrent writers, **437
  reads were rejected** with `"Too Many Requests!"`, and a 64-writer run aborted
  during *initialisation* — N concurrent syncs tripped the limit before any load
  started. A sync is idempotent, so waiting is the correct response; failing fast
  converts a queueing signal into a user-visible error.
- Transient failures do not always look transient. Alongside the expected
  cancellation rendering (`code: 'The operation was cancelled', message: "Timeout
  expired"`), the node produced `code: 'Unknown error', message: "connection
  error: desc = \"i/o timeout\""` — reported as gRPC `Unknown`, so a
  code-only classifier cannot recognise it, and only a transport-level signal
  can. Both renderings belong in the fixture corpus (§2.6).

## §2 — Conventions to follow (established by PR #362)

Reuse these; do not invent a second style. Each is cited so a reviewer can diff
against the precedent.

**1. Config map, optional, defaults preserve today's behavior.**
Rust: a `#[derive(Clone, Debug, Default, PartialEq, Eq)]` config struct with a
nested retry-policy struct, `#[must_use]` `with_*` builders, named getters, and
validated setters returning `Result` (`crates/miden-multisig-client/src/prover.rs`
in #362). TS: an interface hanging off the existing client options object next to
`prover` (`packages/miden-multisig-client/src/prover/config.ts`).

```rust
let rpc = RpcConfig::new()
    .with_url("https://rpc.example")?               // validated: absolute http(s)
    .with_retry_policy(RpcRetryPolicy::new(3));

let client = MultisigClient::builder()
    .miden_endpoint(Endpoint::devnet())
    .guardian_endpoint(guardian_endpoint)
    .rpc_config(rpc)
    .build()
    .await?;
```

```ts
const client = new MultisigClient(midenClient, {
  guardianEndpoint,
  midenRpcEndpoint,
  prover: { retry: { maxAttempts: 4 } },   // #362
  rpc: { retry: { maxAttempts: 3 } },      // this issue
});
```

**2. Normalize and validate at construction, never at first use.**
`max_attempts.max(1)` so a budget always runs at least one attempt; TS
additionally range-checks integer / `0..=u32::MAX` and throws with the field path
in the message (`normalizeMaxAttempts`, `normalizeUrl` in `prover/config.ts`).
URLs must be absolute HTTP(S) with a host — the same rule as `parse_prover_url`,
and the same rule `crates/server/src/evm/config.rs` already applies server-side.

**3. Resolution is a pure, separately tested step.**
`ProverConfig::resolve(default) -> ProverSelection` is a pure function returning
an enum, unit-testable with no network. Mirror it:
`RpcConfig::resolve(default_for_network) -> RpcSelection`.

**4. Injectable runtime so retry tests are deterministic and instant.**
#362's `RetryRuntime` trait (`sleep` + `unit_random`), with a
`ProductionRetryRuntime` and a test implementation. Use the same trait for RPC
retries: exact delay assertions rather than tolerance bands, and no real sleeping
in the suite.

**5. Structured evidence beats substring matching, and permanent wins.**
`StructuredEvidence::{Transient, Permanent, Indeterminate}`, an exhaustive
`tonic::Code` match, the transient HTTP set `{408, 429, 502, 503, 504}`, a
flattened-string fallback only as a last resort, and — the important rule —
permanent evidence anywhere in the error chain overrides transient evidence
anywhere else. The RPC layer should call the same classifier, extended with the
node's rate-limit rendering (`"Too Many Requests!"`) and the transport-level
`i/o timeout` case described above.

**6. A cross-SDK fixture corpus is how parity is enforced.**
#362 puts shared cases in
`speckit/features/001-prover-retry-policy/contracts/prover-policy-fixtures.json`
and reads that same file from both the Rust and the TypeScript tests (attempt
budgets, endpoints, classifications, delays). Ship
`speckit/features/<n>-miden-rpc-policy/contracts/rpc-policy-fixtures.json` the
same way — it is the only mechanism that makes "no silent Rust/TS drift"
(`AGENTS.md` §3 rule 5) mechanically checkable. Run the speckit flow for this
feature, as #360 did.

**7. No silent fallback.** In #362 a custom prover never falls back to the
default or local prover, and the final upstream error is returned unchanged once
attempts are exhausted. Same for RPC: a configured endpoint never silently falls
back to the network default. (§5's failover is the explicit exception, opt-in by
configuring more than one endpoint.)

**8. Docs shape.** A `docs/MULTISIG_SDK.md` subsection with both language
snippets, an explicit statement of what is *not* retried, and a sentence noting
that a larger budget does not add capacity. Correct #362's line "This policy does
not alter or retry Miden RPC requests" — after this work it is false.

## §3 — Custom RPC endpoint (server)

- `GUARDIAN_MIDEN_RPC_ENDPOINT` — when set, replaces
  `NetworkType::rpc_endpoint()`. Validated as an absolute http/https URL at
  startup the way `evm/config.rs` validates its URLs; an invalid value fails
  startup.
- **`NetworkType` keeps its identity role.** It is not only a URL: it feeds the
  dashboard's network-dependent rendering
  (`crates/server/src/dashboard/config.rs:40`), startup logging
  (`crates/server/src/builder/startup.rs:76`), and network config used for
  account-ID prefixes. Overriding the *transport* must not reinterpret the
  *network* — that confusion previously produced the wrong account-ID prefix.
  Keep `GUARDIAN_NETWORK_TYPE` required, treat the URL as an override on top of
  it, and log both resolved values at startup.
- Decide whether a URL override against `MidenTestnet` / `MidenDevnet` should
  warn: pointing a public-network identity at a private node is legitimate for a
  mirror and a mistake otherwise.
- Policy vars mirror the SDK config map: `GUARDIAN_MIDEN_RPC_MAX_ATTEMPTS`,
  `GUARDIAN_MIDEN_RPC_TIMEOUT_MS`.

## §4 — Retry idempotent RPC calls

`crates/miden-rpc-client` exposes seven calls; six are idempotent reads:

| Call | Idempotent | Retry |
|---|---|---|
| `get_status` (`lib.rs:60`) | yes | yes |
| `get_block_header` (`:71`) | yes | yes |
| `sync_state` (`:106`) | yes | yes |
| `get_notes_by_id` (`:136`) | yes | yes |
| `get_account_commitment` (`:159`) | yes | yes |
| `get_account_details` (`:201`) | yes | yes |
| `submit_transaction` (`:91`) | **no** | never — an explicit no-retry arm, with a comment saying why |

- Retry the six reads at the `miden-rpc-client` layer, so the server and every
  other consumer inherit it, under §2's policy — with `max_attempts = 1` (off) as
  the conservative server default until §5 lands.
- SDK: expose read backoff to consumers so a `sync()` against a rate-limited node
  waits instead of failing, per the measurements above.
- TS parity, verified through the fixture corpus (§2.6).

### The prior decision this narrows

Server-side RPC retries were deliberately left out when the prover retry was
designed, for good reasons: canonicalization already retries structurally every
~10s and an RPC error does not consume its retry budget, so a lower-level retry
would add latency inside a lease-holding pass and could mask a real outage; and
request handlers should fail fast rather than hold a connection and a lock, which
matters most at the read concurrency #317 targets.

That reasoning still holds **for the canonicalization pass and for request
handlers**, and this issue does not overturn it. What it reopens is narrower:

- retry at the **connect/startup** boundary, which holds no lease and no lock, so
  a node blip at boot need not fail startup;
- an **opt-in, default-off** read retry for deployments sitting behind a
  rate-limited public RPC, where fail-fast turns a queueing signal into a
  user-visible error;
- with §5, a failed read triggers **failover**, which is a different action from
  waiting.

Anything enabled inside the canonicalization pass must be measured against the
lease duration before it ships, not assumed safe.

## §5 — Multiple RPC endpoints with failover (design now, build later)

Single-endpoint configuration is a hard availability ceiling: if the configured
node is down, rate-limiting, or lagging, Guardian's verification path stops — and
today there is not even a way to repoint it without a redeploy. Direction,
recorded here so §3/§4 don't have to be redone:

- `GUARDIAN_MIDEN_RPC_ENDPOINTS` — an ordered list; §3's single-value var becomes
  the one-element case rather than a separate mechanism.
- Fail over on transport errors, `Unavailable`, and rate-limit responses — never
  on `Permanent` evidence (§2.5), which every honest node returns identically.
  The classifier already draws exactly this line, which is why it should be
  shared.
- Endpoints must be **same-network peers**. Failing over to a node on a different
  network would silently verify against the wrong chain — a correctness and
  custody problem, not a degraded-service one. Verify at startup that every
  configured endpoint reports the same network/genesis, and refuse to start
  otherwise.
- State the retry/failover interaction: retry-then-failover, or failover-first?
  Spending the whole budget on a saturated node before trying a healthy peer is
  waste.
- Nodes lag differently: a commitment read from a lagging peer reports "not yet on
  chain" for a delta that has landed. The canonicalization path must tolerate a
  mid-decision peer switch, or pin one endpoint per pass.
- Per-endpoint metric labels, so an operator can see which endpoint served what
  and how often failover fired.

Not in the acceptance criteria — this needs its own issue once §3/§4 fix the
configuration surface.

## Non-goals

- Prover retry configurability — #360 / PR #362.
- Retrying `submit_transaction`, or retrying the surrounding transaction
  execution. The delta reaches Guardian for its acknowledgment before proving
  starts, so re-executing re-pushes that delta and is refused as a pending-delta
  conflict; #362's transaction-safety section covers the same boundary.
- Changing default behavior. Everything here is opt-in, with today's values as
  defaults.

## Open questions

- Should the server's read retry default on or off? Off is safer given the
  fail-fast reasoning above; a deployment against public testnet is materially
  more reliable with it on. Possibly: on for reads outside the canonicalization
  lease, off inside it.
- Does a URL override combined with a public-network `NetworkType` need to be
  blocked, warned about, or allowed silently?
- Should `GUARDIAN_MIDEN_RPC_ENDPOINT` accept a bare `host:port` (matching
  `NetworkType`'s current forms) or require a scheme? Requiring a scheme matches
  `evm/config.rs`'s validation and avoids TLS-by-accident.
- Where should the shared classifier live? #362 puts it in
  `crates/miden-multisig-client`, but the server needs it too and does not depend
  on the multisig SDK — so it likely belongs in a lower crate that both
  `miden-multisig-client` and `miden-rpc-client` can depend on.

zeljkoX:
Adds configurable timeout and opt-in retries for Miden node RPC across the server and both multisig SDKs, with one shared policy core. Closes #364.

> [!IMPORTANT]
> **Upgrade note:** the default per-request Miden RPC deadline decreases from 30 seconds to 10 seconds for existing deployments. The deadline is applied at the tonic channel level, so it caps submissions as well as reads even though submissions remain single-attempt. Set `GUARDIAN_MIDEN_RPC_TIMEOUT_MS=30000` to retain the previous server deadline.

## RPC configuration

- **Shared core** (`guardian-shared::retry`): attempt budget, jittered exponential backoff, and transient/permanent error classification (permanent evidence anywhere vetoes transient anywhere), with one generic retry loop used by every Rust retry site. The TypeScript SDK mirrors the classifier and loop; behavior is pinned across SDKs by a fixture corpus (classification verdicts + exact delays).
- **Rust SDK**: `RpcConfig` on the builder (`with_timeout_ms`, `with_retry_policy`) wraps the node client and the private-note transport. Reads default to two total attempts; `RpcRetryPolicy::new(1)` opts out. miden-client's internal transport-retry loop is disabled (`with_max_retries(0)`) — it silently retransmitted rate-limited requests **including submissions**. Submissions and note sends are never retried, under any configuration.
- **TypeScript SDK**: `rpc: { retry: { maxAttempts } }` wraps the SDK's direct node reads. No timeout by design — the browser WASM client cannot cancel an in-flight request.

## Note transport endpoint

Private-note relay uses a transport service separate from the node RPC. The testnet/devnet presets derive it automatically; custom node endpoints previously had relay silently disabled. New optional `note_transport_endpoint()` on the Rust builder enables it explicitly (TS sets `noteTransportUrl` on the injected `MidenClient`). The transport shares the RPC resilience policy: fetches and stream establishment retry, sends never retry in-call.

## Server

- `GUARDIAN_MIDEN_RPC_ENDPOINT` / `_TIMEOUT_MS` (default 10000) / `_MAX_ATTEMPTS` (default 1 = retries off) via a network-owned `RpcSettings` abstraction; the builder stays network-agnostic and rejects settings resolved for a different network. Endpoint overrides must be origin-only (`scheme://host[:port]`); embedded credentials, paths, queries, and fragments are rejected because tonic does not send them as authentication.
- Canonicalization reads stay single-attempt (`RpcReadMode::SingleAttempt`) — its scheduled passes are the retry; the abandon-candidate service uses the configured policy.
- Retry activity surfaces as `guardian_miden_rpc_retries_total` (per operation).
- Every Miden RPC surface now shares the same 10s default per-request deadline.

## Testing

Unit suites per wrapper, the cross-SDK fixture corpus, and a new wire-level tier: a scripted in-process gRPC node (`miden-rpc-client/src/test_node.rs`) exercising real transport renderings — 429 recovery, rate-limit text, permanent fail-fast, deadline expiry, dropped connections — plus a guard asserting the miden-client internal retry loop stays disabled (2 configured attempts must reach the node as exactly 2 requests; mutation-verified).

## Summary by CodeRabbit

* **New Features**
  * Added configurable Miden RPC endpoints, timeouts, retry limits, and note-transport endpoints.
  * Added automatic retries for transient, read-only RPC and note-fetch operations, while keeping submissions single-attempt.
  * Added RPC retry metrics and configuration support across server, Rust SDK, TypeScript SDK, and examples.
* **Bug Fixes**
  * Improved error context and preserved underlying RPC and transaction failure details.
  * Prevented unnecessary retries during canonicalization verification.
* **Documentation**
  * Expanded configuration, setup, troubleshooting, and SDK guidance for RPC behavior.


haseebrabbani (approved):
LGTM — one release-note callout: the default per-request deadline drops 30s→10s for existing deployments (channel-level, so it caps submissions too).

Minor follow-ups: `Display` on the new `*_with_context` errors loses the underlying cause, and the docs' "may embed credentials" claim isn't honored by tonic (userinfo is redacted but never sent as auth).

zeljkoX:
> LGTM — one release-note callout: the default per-request deadline drops 30s→10s for existing deployments (channel-level, so it caps submissions too).
> 
> Minor follow-ups: `Display` on the new `*_with_context` errors loses the underlying cause, and the docs' "may embed credentials" claim isn't honored by tonic (userinfo is redacted but never sent as auth).

Thanks for feedback. Applied it. Will extend release notes to denote this change.

PR #375, opened 2026-08-04
fix(server): reconstruct first-tx SwitchGuardian with the guardian selector re-enabled

### LINKED ISSUE #369: Chain-reconciliation sweep for guardian switches that never reach the push path (release-on-switch v2)
## Problem

#311 (shipped in v0.15.1) added release-on-switch detection, but it is **push-triggered only**: it runs when the pre-switch guardian commits/canonicalizes the `SwitchGuardian` delta that the switching client pushes via `execute_proposal` (client-side since #287 / v0.15.0). The hook fires on exactly two paths — optimistic delta commit (`services/delta_commit.rs`) and candidate canonicalization (`jobs/canonicalization/processor.rs`) — and there is no other trigger: no startup scan, no periodic reconciliation.

That leaves permanent blind spots where the old guardian keeps an account listed as active (`released_at: null`) even though it verifiably lost it on-chain:

1. **Switches that predate the guardian running ≥ v0.15.1** — nothing ever revisits them.
2. **Clients older than v0.15.0** — they never push the switch delta at all.
3. **Failed pushes** — the push is best-effort by design; both the Rust and TS clients swallow errors silently (`catch {}` in `executeProposal`) with no retry.
4. **The offline switch path** (`ProposalResult::Offline`) — exists by design and never pushes.

## Proposal

A periodic reconciliation job (or an extension of the existing canonicalization worker loop) that, for each non-released, non-EVM account:

1. Compares the guardian's stored `current_commitment` against chain via the existing `NetworkClient::verify_commitment` (already returns `StateVerification::Mismatch { on_chain }`).
2. On sustained mismatch with no pending candidate that explains it, classifies the account as having moved without this guardian's knowledge — deltas are prev-commitment-bound, so this server's state can never advance again.
3. Releases via the existing `set_released` path (same terminal semantics, audit event `accounts.release` with a distinct `detected_by: "reconciliation_sweep"` payload marker vs. the push-path detection).

Reuse the confirmation-count pattern from #312 (`divergence_confirmations`) so a transient RPC anomaly or in-flight canonicalization can't release an account spuriously.

## Open questions

- **Classification for private accounts.** For public accounts the sweep can fetch on-chain state and extract the guardian key slot (`extract_guardian_commitment`) to positively confirm a switch. For private accounts only the commitment is on-chain — a sustained mismatch proves this server's state is dead-ended, but not *why* (switch vs. missed delta while still being the guardian). Options: release on dead-end regardless (the account is unusable through this server either way; `/configure` re-onboarding heals false positives), or introduce a distinct non-terminal state for "diverged, cause unknown".
- **Interplay with the #312 divergence machinery** — the sweep must not race candidate canonicalization or the diverged-candidate release path.
- **RPC load / pacing** for large account sets (OZ instance: ~1,900 accounts); batching and a sweep interval config.
- **v3 alternative**: a node-side filtered events API (upstream Miden node) would make this event-driven instead of polled — the sweep is the interim that works against today's node.

## Related

- #305 (original split-brain issue), #311 (push-triggered detection), #287 (client-side push), #312 (diverged-candidate release)

haseebrabbani:
## Problem

Release-on-switch (#305 / #311) never fires for an account whose **first transaction** is a `SwitchGuardian`, leaving the pre-switch guardian serving the account forever — the exact split-brain #305 was opened for. Reported live on the OpenZeppelin Guardian v0.16.0 (see the discussion on #369): switched-away accounts still `released_at: null`.

Root cause: a first transaction pushes a **full-state** delta, and `apply_delta` took that delta's guardian **selector** slot as authoritative. But a `SwitchGuardian` abort summary captures the selector mid-script (OFF — it is disabled during the switch), while the guardian-verified execution ends with `enable_guardian`, so the chain's post state has it ON. The reconstructed commitment therefore never matches the chain, the canonicalization worker condemns the candidate as **diverged** (~2 ticks), discards it together with its proposal, and the release hook — which only runs on successful canonicalization — never executes.

The partial-delta path already handles this with a pre-tx-gated re-enable; the full-state path skipped it on the assumption that "the full-state delta carries the authoritative selector", which is wrong for exactly this transaction type.

## Fix

Derive `guardian_enabled_pre_tx` for the full-state path from `prev_state_json` — the initial state registered via `/configure` — mirroring the partial-path rule. An unparseable prev state falls back to the delta's own selector (previous behavior). Regression test `test_apply_delta_first_tx_switch_reenables_selector` pins both directions: guardian-enabled initial state → selector re-enabled; guardian-disabled initial state → delta's OFF selector preserved.

Second commit: both clients previously discarded the best-effort switch-push error entirely (Rust `let _ =`, TS `catch {}`). A lost push produces this same split-brain with no trace anywhere to diagnose it by — the failure is still non-blocking, but now logged (`tracing::warn` / `console.warn`).

## Notes

- This does not close #369: the reconciliation sweep is still needed for pushes that are lost before reaching the server, for pre-#311 switches, and for clients that never push.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved guardian-switch handling for first transactions, preserving the correct authorization state when guardian protection is enabled or disabled.
  * Guardian-switch execution now continues safely when optional state updates or canonicalization fail.
* **Improvements**
  * Added warning logs when non-blocking guardian-switch operations encounter errors, making issues easier to identify without interrupting execution.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4788,6 +4788,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "tonic",
+ "tracing",
  "url",
 ]
 
--- a/crates/miden-multisig-client/Cargo.toml
+++ b/crates/miden-multisig-client/Cargo.toml
@@ -35,6 +35,9 @@ base64 = { workspace = true }
 
 # Error handling
 thiserror = { workspace = true }
+
+# Logging (facade only; host applications install the subscriber)
+tracing = "0.1"
 anyhow = { workspace = true }
 
 # Utilities
--- a/crates/miden-multisig-client/src/client/proposals.rs
+++ b/crates/miden-multisig-client/src/client/proposals.rs
@@ -263,14 +263,28 @@ impl MultisigClient {
                 .await?;
             signature_advice.push(guardian_advice);
         } else {
-            let _ = self
+            // SwitchGuardian: push the delta to the pre-switch GUARDIAN so it
+            // canonicalizes there and the account is released (issue #305).
+            // Best-effort — an unreachable GUARDIAN must not block the switch —
+            // but the outcome must be observable: a silently lost push leaves
+            // the old GUARDIAN serving a released account (split-brain) with
+            // nothing in any log to diagnose it by.
+            if let Err(error) = self
                 .get_guardian_ack_signature(
                     &account,
                     proposal.nonce,
                     &proposal.tx_summary,
                     tx_summary_commitment,
                 )
-                .await;
+                .await
+            {
+                tracing::warn!(
+                    %error,
+                    "best-effort SwitchGuardian delta push to the pre-switch \
+                     GUARDIAN failed; it will keep serving this account until \
+                     reconciliation"
+                );
+            }
         }
 
         // Build the final transaction request with all signatures
--- a/crates/server/src/network/miden/mod.rs
+++ b/crates/server/src/network/miden/mod.rs
@@ -194,13 +194,25 @@ impl NetworkClient for MidenNetworkClient {
         let account_delta = tx_summary.account_delta();
 
         // Check if this is a full state delta (new account deployment) or partial delta (update)
-        let mut guardian_enabled_pre_tx = false;
+        let guardian_enabled_pre_tx;
         let mut account = if account_delta.is_full_state() {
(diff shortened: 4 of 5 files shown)


zeljkoX (approved):
LGTM

PR #379, opened 2026-08-05
Improve CONCEPTS.md

zeljkoX:
Improve CONCEPTS.md docs. Clarify Guardian ACK.

## Summary by CodeRabbit

- **Documentation**
  - Clarified that Guardians act as a separate, removable authentication gate rather than multisig signers.
  - Documented that user keys satisfy the M-of-N threshold, while ordinary transactions also require a Guardian acknowledgment.
  - Clarified that Guardians cannot authorize transactions independently and that Guardian changes are exempt from acknowledgment checks.
  - Updated ACK definitions and documented Guardian rotation through the user-authorized `SwitchGuardian` transaction.

### DIFF
--- a/docs/CONCEPTS.md
+++ b/docs/CONCEPTS.md
@@ -30,26 +30,45 @@ Guardian is:
 ## The custody spectrum
 
 Traditional crypto custody is binary — either a full custodian holds the
-key, or the user does. Guardian creates a third position by serving as one
-signer in a multisig arrangement, typically:
+key, or the user does. Guardian creates a third position — but **not** by
+holding a seat in the user's multisig. The account component enforces two
+independent checks on every transaction:
 
 ```mermaid
 flowchart LR
-  H["User hot key<br/>(daily transactions)"]
-  C["User cold key<br/>(recovery, offline)"]
-  G["Guardian service key<br/>(policy / co-signing)"]
-  T["2-of-3 threshold"]
+  subgraph S["User signer set"]
+    H["Hot key<br/>(daily transactions)"]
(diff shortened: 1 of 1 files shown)


haseebrabbani (approved):
Minor suggestions:

- (custody spectrum, rotation exception bullet): "The one transaction the account component executes without the Guardian check" is true for guardian-enabled accounts, but the guardian selector slot can also be OFF entirely (e.g. around first-tx switch reconstruction). Maybe qualify: "For a guardian-enabled account, the one transaction…"
- (same bullet / provider rotation): "needs only the user threshold" — worth a nod to per-procedure threshold overrides, since an override on update_guardian (potentially 1) changes the effective bar for removing the Guardian's veto. The "applicable user threshold" wording later in the doc already covers this; mirroring it here would keep the two sections aligned.

zeljkoX:
> Minor suggestions:
> 
> * (custody spectrum, rotation exception bullet): "The one transaction the account component executes without the Guardian check" is true for guardian-enabled accounts, but the guardian selector slot can also be OFF entirely (e.g. around first-tx switch reconstruction). Maybe qualify: "For a guardian-enabled account, the one transaction…"
> * (same bullet / provider rotation): "needs only the user threshold" — worth a nod to per-procedure threshold overrides, since an override on update_guardian (potentially 1) changes the effective bar for removing the Guardian's veto. The "applicable user threshold" wording later in the doc already covers this; mirroring it here would keep the two sections aligned.

Upstream contracts do not have guardian enabled flag and we are soon migrating so I would keep it as is due to that.
Thanks for the feedback.

PR #380, opened 2026-08-06
fix(server): validate cosigner_commitments against initial account state

### LINKED ISSUE #102: Validate cosigner_commitments on configure
`/configure` accepts` auth.cosigner_commitments` from the client and only verifies that the signer’s pubkey exists in the account state. The full list isn’t validated against the actual account storage, so incorrect or malicious auth config could be stored.

When configuring an account, extract signer commitments from `initial_state` and ensure they match `auth.cosigner_commitments` (same set/order as expected), otherwise reject

haseebrabbani:
Closes #102. Supersedes #109, which predates the `GuardianError` migration and no longer applies.

## Problem

`/configure` accepted the client-declared `auth.cosigner_commitments` after only verifying that the *requesting* key exists in the account state (`validate_credential`). The rest of the list was stored unvalidated — and that stored list is the authorization source of truth for every subsequent request (`Auth::verify` authorizes any key whose commitment appears in it). A configure call could therefore register commitments that are not signers of the account.

## Fix

After the guardian-binding check, extract the signer set from `initial_state` (via the existing `NetworkClient::should_update_auth`, which reads the `openzeppelin::multisig::signer_public_keys` map) and reject with `InvalidInput` when the declared list does not match it exactly (same set and order — the SDK derives the list from the same map walk, so honest clients are unaffected).

`Ok(None)` (no extractable signer set) skips the check, matching the canonicalization-side semantics of `should_update_auth`. Every real multisig-guardian account carries a non-empty signer map — `MultisigGuardianConfig` writes all signers, including 1-of-1 — so the check always runs for real accounts.

## Summary by CodeRabbit

* **New Features**
  * Added access to all authorized cosigner commitments for supported authentication methods.

* **Bug Fixes**
  * Account configuration now rejects declared signer sets that do not match the account’s actual signer commitments.
  * Configuration failures are reported when signer commitments cannot be extracted.
  * Multisignature onboarding and guardian-switching flows now validate the complete cosigner set.

### DIFF
--- a/crates/server/src/metadata/auth/mod.rs
+++ b/crates/server/src/metadata/auth/mod.rs
@@ -77,6 +77,20 @@ impl Auth {
         }
     }
 
+    /// The per-scheme list of authorized signer commitments
+    /// (`signers` for EVM accounts).
+    pub fn cosigner_commitments(&self) -> &[String] {
+        match self {
+            Auth::MidenFalconRpo {
+                cosigner_commitments,
+            }
+            | Auth::MidenEcdsa {
+                cosigner_commitments,
+            } => cosigner_commitments,
+            Auth::EvmEcdsa { signers } => signers,
+        }
+    }
+
     pub fn with_updated_commitments(&self, cosigner_commitments: Vec<String>) -> Self {
         match self {
             Auth::MidenFalconRpo { .. } => Auth::MidenFalconRpo {
--- a/crates/server/src/services/configure_account.rs
+++ b/crates/server/src/services/configure_account.rs
@@ -84,6 +84,42 @@ pub async fn configure_account(
                 ))
             })?;
 
+        // The stored cosigner list is the authorization source of truth for
+        // every later request (`Auth::verify`), so the full client-declared
+        // set must match the signer set actually stored in the submitted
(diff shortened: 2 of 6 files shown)


zeljkoX (approved):

Looks good.

Few nits:
- `test_configure_account_with_real_miden_account` still uses a mock path where should_update_auth() returns Ok(None). Consequently, it succeeds without proving that the supplied commitments match the account state and would continue passing if the new validation were removed.
- Update `spec/processes.md` to state that, for MultisigGuardian accounts, auth.cosigner_commitments must exactly match the signer map extracted from initial_state, including its canonical order. This is a user-visible auth behavior change and helps explain otherwise confusing InvalidInput responses.
- Add `test_rejects_reordered_commitments` to lock in the deliberately order-sensitive behavior.
- Consider validating commitments at ingestion as canonical 0x-prefixed, lowercase, 32-byte hex and explicitly rejecting empty or duplicate lists. Current exact string equality works with the SDKs but can produce confusing failures for manual clients.
- Consider mapping malformed client-supplied initial_state to InvalidInput rather than NetworkError. This is consistent with existing behavior, so it need not block this PR.

PR #381, opened 2026-08-06
feat: support P2IDE reclaim/timelock heights in send proposals (#366)

### LINKED ISSUE #366: multisig-client: createP2idProposal should support reclaim/timelock heights (P2IDE)
## Context

`@openzeppelin/miden-multisig-client`'s `createP2idProposal` only accepts `(recipientId, faucetId, amount, nonce?, { noteType? })` — there is no way to set a reclaim height (or timelock height), so every multisig send builds a plain P2ID note.

In the Miden wallet, sends carry a user-picked expiration date that maps to a P2IDE reclaim height (`reclaimAfter`). On Guardian accounts that value was silently dropped by the proposal layer, making every guardian send permanently non-recallable — the wallet UI showed an expiration that never reached the chain (wallet-side issue: 0xMiden/wallet#308 has the P2IDE background).

## Request

Extend the send proposal API to carry the P2IDE heights, e.g.:

```ts
createP2idProposal(recipientId, faucetId, amount, nonce?, {
  noteType?: NoteType;
  reclaimHeight?: number;   // absolute block height; presence selects P2IDE over P2ID
  timelockHeight?: number;  // optional consume-not-before height
})
```

- Presence of either height should build a P2IDE note instead of P2ID (mirroring the SDK's `reclaimAfter`/`timelockUntil` semantics on `SendOptions`).
- The guardian server should treat these like any other send proposal (no new proposal type needed).

haseebrabbani:
Closes #366.

`createP2idProposal` only built plain P2ID notes, so the wallet-selected expiration
(`reclaimAfter`) was silently dropped by the proposal layer and every guardian send was
permanently non-recallable (wallet-side background: 0xMiden/wallet#308).

## What changed

- **TS client**: `createP2idProposal(recipient, faucet, amount, nonce?, { noteType, reclaimHeight, timelockHeight })`.
  Presence of either height builds a **P2IDE** note (`NoteScript.p2ide()`, storage
  `[recipient_suffix, recipient_prefix, reclaim, timelock]` with `0` = unset, per
  miden-standards `P2ideNoteStorage`) instead of plain P2ID. The heights are recorded in
  proposal metadata, and the rebuild-from-metadata, execution, and `getP2idNoteId` paths
  thread them back so cosigners verify and execute the identical note. The serial number
  stays deterministic from the salt, so the note ID is still known ahead of execution.
- **Rust client**: `TransactionType::P2ID` carries `reclaim_height`/`timelock_height`
  (`transfer()`/`transfer_with_note_type()` default to `None`; new `transfer_p2ide()` for
  explicit heights). `build_p2id_transaction_request` builds a `P2ideNote` when either
  height is set, drawing the serial from the same salt-seeded rng. `ProposalMetadata`,
  `ProposalMetadataPayload`, and `ExportedMetadata` round-trip the fields.
- **Wire format**: `metadata.reclaim_height` / `metadata.timelock_height` are `u32` block
  heights, omitted when unset so plain-P2ID payloads keep the pre-#366 wire shape; absent
  still means plain P2ID, so pre-existing proposals rebuild identically. `0` is rejected at
  creation and parse time (it is the on-chain encoding for "no constraint"), and invalid
  values fail closed instead of silently rebuilding a note that could never match the
  signed tx_summary commitment — same fail-closed posture as `note_type` (#322).
- **Server + operator client**: the fields are mirrored in the delta-summary proposal
  metadata for the dashboard; OpenAPI specs regenerated. The server treats these as
  ordinary send proposals — no new proposal type.
- **Examples/docs**: demo CLI prompts for optional reclaim/timelock heights; browser
  example wrappers pass the heights through; `MULTISIG_SDK.md`, `spec/api.md`, and both
  package READMEs updated.

## Compatibility

- Pre-#366 proposals (no height fields) rebuild byte-identically as plain P2ID.
- Older clients receiving a P2IDE proposal fail closed with a commitment mismatch rather
  than executing a wrong note — the same tradeoff accepted in #322/#229.
- Release order: `@openzeppelin/guardian-client` must publish before
  `@openzeppelin/miden-multisig-client` picks up the new wire fields (same as #322).

## New Features
- Added optional reclaim and timelock block heights for P2IDE transfers.
- P2ID proposals now preserve these settings through creation, signing, serialization, and reconstruction.
- Added validation for invalid, zero, and out-of-range heights.
- Updated browser, TypeScript, Rust, demo, and API interfaces to support the new options.

## Documentation
- Added usage examples and API documentation for reclaimable and timelocked transfers.


crates/miden-multisig-client/src/client/notes.rs
```diff
@@ -271,17 +274,36 @@ impl MultisigClient {
...
-            *recipient,
-            vec![asset.into()],
-            *note_type,
-            Default::default(),
-            &mut rng,
-        )
-        .map_err(|e| {
-            MultisigError::TransactionExecution(format!("failed to build P2ID note: {}", e))
-        })?;
+        let note = if reclaim_height.is_some() || timelock_height.is_some() {
```
zeljkoX: Let's do something similar like in  Rust version in order to improve readability where we have separate variable to store is it p2ide value to and use it afterwards in if conditions and where needed.

crates/miden-multisig-client/README.md
```diff
@@ -69,6 +69,14 @@ let tx = TransactionType::transfer(recipient, faucet, 1_000);
 use miden_protocol::note::NoteType;
 let tx = TransactionType::transfer_with_note_type(recipient, faucet, 1_000, NoteType::Private);
 
+// For a reclaimable and/or timelocked send use `transfer_p2ide` (issue #366):
+// presence of either height creates a P2IDE note instead of a plain P2ID note.
+let tx = TransactionType::transfer_p2ide(
+    recipient, faucet, 1_000, NoteType::Public,
```
zeljkoX, line 160: Does it make sense to use new line for every parameter or this is because readability, new parameters are in new lines?

crates/server/src/delta_summary/mod.rs
```diff
@@ -144,6 +144,21 @@ pub struct ProposalMetadata {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub note_type: Option<String>,
 
+    /// P2IDE reclaim block height (issue #366). Presence of either height
+    /// means the proposal creates a P2IDE note; absent => plain P2ID.
+    /// Valid heights are 1..=u32::MAX; 0 encodes "no constraint" on-chain
+    /// and is rejected by every client parser.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[schema(minimum = 1, maximum = 4294967295_u32)]
```
zeljkoX: Let's enforce with serde documented parameters values.

crates/server/src/delta_summary/mod.rs
```diff
@@ -144,6 +144,21 @@ pub struct ProposalMetadata {
...
+    pub reclaim_height: Option<u32>,
+
+    /// P2IDE timelock block height (issue #366).
+    /// Valid heights are 1..=u32::MAX; 0 encodes "no constraint" on-chain
+    /// and is rejected by every client parser.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[schema(minimum = 1, maximum = 4294967295_u32)]
```
zeljkoX: +1

docs/MULTISIG_SDK.md
```diff
@@ -437,8 +437,26 @@ const privateProposal = await multisig.createP2idProposal(
...
+// P2IDE note (issue #366): pass `reclaimHeight` and/or `timelockHeight`
+// (absolute block heights) to create a reclaimable/timelocked note instead
+// of a plain P2ID note. `reclaimHeight` lets the sender reclaim the note if
+// it is still unconsumed at that block; `timelockHeight` prevents
+// consumption before that block.
+const reclaimableProposal = await multisig.createP2idProposal(
+  recipientAccountId,
+  faucetAccountId,
+  1000n,
+  undefined,
```
zeljkoX, line 572: I know it is not part of PR but this example looks bad due to usage of undefined.
I am in favor of moving nonce to options so API would be cleaner.
I guess for start we would need to have logic to detect 4th param type.

It is not required for this PR... If not addressed with this one let's create GH issue.

docs/MULTISIG_SDK.md
```diff
@@ -733,6 +751,15 @@ let tx = TransactionType::transfer_with_note_type(
     recipient_id, faucet_id, 1000, NoteType::Private,
 );
 
+// P2IDE Transfer (issue #366): optional reclaim and/or timelock block
+// heights. Presence of either height creates a P2IDE note instead of a
+// plain P2ID note.
+let tx = TransactionType::transfer_p2ide(
+    recipient_id, faucet_id, 1000, NoteType::Public,
+    Some(500_000), // reclaim_height
```
zeljkoX: The fact that this example needs  `reclaim_height` `timelock_height` comments to be readable is a hint the signature is swap-prone: two adjacent Option<u32> params, and the compiler can't catch a caller mixing them up. A swapped pair still builds a perfectly consistent, signed, verifiable note, just with the sender able to reclaim before the recipient's timelock ends.

Suggestion: take a small named struct instead, with NonZeroU32 so the zero-height case becomes unrepresentable and the runtime 0 checks (validate_p2ide_height, ProposalMetadata::p2ide_height) can be dropped, since serde rejects 0 for NonZeroU32 at parse time:

```
 pub struct P2ideHeights {
     pub reclaim: Option<NonZeroU32>,
     pub timelock: Option<NonZeroU32>,
 }

 let tx = TransactionType::transfer_p2ide(
     recipient, faucet_id, 1_000, NoteType::Public,
     P2ideHeights { reclaim: NonZeroU32::new(500_000), timelock: None },
 );
```

zeljkoX (commented):
LGTM

Added few comments.

This PR introduces changes that could affect client consumers and their logic due to new params on Rust crates side.

For now our consumers are mostly using npm packages.

Could you review this and decide what version should we use this on release, should we bump to 0.17.0 due to changes or it is acceptable to proceed with 0.16.x?

docs/MULTISIG_SDK.md
haseebrabbani: created issue https://github.com/OpenZeppelin/guardian/issues/387

haseebrabbani:
@zeljkoX I recommend **0.17.0** for the release version.

The Rust changes are breaking: `TransactionType::P2ID` changed shape (new `heights: P2ideHeights` field), and `transfer_p2ide` / `with_payment_metadata` / `build_p2id_transaction_request` changed signatures. Cargo treats `0.16.x → 0.16.y` as compatible, so shipping this as 0.16.x would let a `cargo update` break Rust consumers silently

The npm packages are strictly additive (new optional fields/options), so npm consumers are unaffected either way; bumping them to 0.17.0 in lockstep is harmless and keeps the versions aligned.

zeljkoX (approved):

PR #383, opened 2026-08-07
fix(contracts): read auth flags from element 0 after the little-endian migration

### LINKED ISSUE #382: Guardian selector and replay guard read the wrong felt after the v0.14 endianness change
The v0.14 upgrade moved these components to Miden's little-endian stack convention and updated the word reads to `movdn.3 drop drop drop`. It looks like four reads were missed:

- `crates/contracts/masm/auth/guardian.masm:182`
- `crates/contracts/masm/auth/guardian_ecdsa.masm:178`
- `crates/contracts/masm/auth/multisig.masm:87`
- `crates/contracts/masm/auth/multisig_ecdsa.masm:87`

These read a word where the value sits in element 0 and the other three are zeros, so a bare `drop drop drop` keeps element 3, which is always 0. 

Leading to two bugs:

1. In `verify_guardian_signature` the selector always reads as 0, so `push.1 eq` is never true and the signature check inside the `if.true` never runs. An account created with `guardian_enabled = true` behaves like a plain multisig.

2. In `assert_new_tx` the previous map value always reads as 0, so `assertz` always passes and `ERR_TX_ALREADY_EXECUTED` can't fire.

ngjupeng:
Closes #382

The v0.14 upgrade (#185) moved these components to Miden's little-endian operand-stack convention and updated twelve word reads to `movdn.3 drop drop drop`. Four reads were missed, so both `verify_guardian_signature` and `assert_new_tx` inspect the padding felt at element 3 instead of the flag at element 0, which is always zero.

As a result the guardian branch in `verify_guardian_signature` is never entered, and `ERR_TX_ALREADY_EXECUTED` in `assert_new_tx` is unreachable.

Applies `movdn.3` at the four sites:

- `crates/contracts/masm/auth/guardian.masm:182`
- `crates/contracts/masm/auth/guardian_ecdsa.masm:178`
- `crates/contracts/masm/auth/multisig.masm:87`
- `crates/contracts/masm/auth/multisig_ecdsa.masm:87`

Also corrects two stale stack comments that described the pre-migration layout, and regenerates the packaged MASM and TS copies.

Adds regression tests for both rejection paths. Neither was previously covered: the existing tests always attach a valid guardian signature and submit each transaction once, so both branches were unreachable in the suite.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved multisignature transaction validation and guardian authentication handling.
  * Prevented valid guardian signatures from being misread during authorization.
  * Strengthened replay protection by correctly preserving transaction execution status.
  * Applied fixes consistently across supported Falcon and ECDSA authentication flows.

* **Tests**
  * Added regression coverage for missing guardian signatures.
  * Added coverage ensuring previously executed transactions cannot be replayed.

### DIFF
--- a/crates/contracts/masm/auth/guardian.masm
+++ b/crates/contracts/masm/auth/guardian.masm
@@ -179,7 +179,7 @@ end
 pub proc verify_guardian_signature(msg: word)
     push.GUARDIAN_SELECTOR_SLOT[0..2]
     exec.active_account::get_item
-    drop drop drop
+    movdn.3 drop drop drop
     # => [selector, MSG]
 
     push.1 eq
--- a/crates/contracts/masm/auth/guardian_ecdsa.masm
+++ b/crates/contracts/masm/auth/guardian_ecdsa.masm
@@ -175,7 +175,7 @@ end
 pub proc verify_guardian_signature(msg: word)
     push.GUARDIAN_SELECTOR_SLOT[0..2]
     exec.active_account::get_item
-    drop drop drop
+    movdn.3 drop drop drop
     # => [selector, MSG]
 
     push.1 eq
--- a/crates/contracts/masm/auth/multisig.masm
+++ b/crates/contracts/masm/auth/multisig.masm
@@ -82,9 +82,9 @@ pub proc assert_new_tx(msg: word)
 
     # Set the key value pair in the map to mark transaction as executed
     exec.native_account::set_map_item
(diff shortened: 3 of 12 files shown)


zeljkoX:
Thanks again for identifying this security bug.

I am closing this issue and its associated PR in favor of merged PR #340.

Looking forward to your future contributions.

Thanks!

PR #390, opened 2026-08-11
feat: 341 Introduce secondary subdomains

zeljkoX:
As part of subdomains migration this PR introduces changes to support old and new subdomains during migration period.

## Summary by CodeRabbit

* **New Features**
  * Added optional legacy hostname support during domain migrations.
  * Supports HTTPS certificates, DNS records, and gRPC endpoints for both canonical and alias hostnames.
  * Deployment output and validation now cover configured secondary hostnames.

* **Documentation**
  * Added configuration guidance for canonical and alias domains, certificates, and DNS providers.
  * Added a domain migration runbook covering deployment safeguards, verification, monitoring, and legacy hostname removal.
  * Updated deployment examples and infrastructure documentation with current hostname settings and migration behavior.


haseebrabbani (approved):
Approving, but would be good to get @collins-w approval as well. 

Remaining comments are more about operator-procedure hardening rather than code issues. Before running the migration (not blocking merge), we should probably fix these in the runbook/config:

1. **Pin the DNS-provider vars in the export blocks** (`CLOUDFLARE_ZONE_ID`, `CLOUDFLARE_API_TOKEN`, `ROUTE53_ZONE_ID`). They're commented out in `.env.example`, and without them both record counts go to 0 — the post-`state mv` plan then proposes destroying the live legacy record, with only plan review in the way.
2. **Tighten the Cloudflare provider constraint to `~> 5.0`** (and ideally commit `.terraform.lock.hcl`): `>= 4.0.0` permits v4, which doesn't have `cloudflare_dns_record`, and the runbook's plan-shape expectations depend on v5 behavior.
3. **Add a `terraform -chdir=infra init` step before the state-move guards** — `state show` needs provider schemas, so on a fresh checkout the guards always refuse with "found <empty>" and invite an unguarded manual `state mv`.

Issue #306, opened 2026-06-27
Guardian auth component unrecognized by miden-standards AccountInterface — getPublicKeyCommitments() returns [] for Guardian accounts


WiktorStarczewski:
### Summary

A Guardian account's authentication component (built by `@openzeppelin/miden-multisig-client`) is not recognized by the Miden SDK's `AccountInterface`, so `Account.getPublicKeyCommitments()` returns `[]` for Guardian accounts. Any consumer that derives an account's public key through the SDK gets nothing back for Guardian accounts, while plain single-sig accounts work fine.

### Environment

- `@openzeppelin/miden-multisig-client@0.15.0-rc.0`
- `@miden-sdk/miden-sdk@0.15.2` (links `miden-standards` / `miden-protocol` `0.15.3`)

### Root cause

`AccountInterface::from_account` (miden-standards) classifies an account's components by **exact MAST-root subset containment** against a fixed set of bundled standard component templates, whose procedures live in the `miden::standards::auth::*` namespace.

A Guardian account's auth component is compiled at runtime from OZ's own MASM (`auth_tx_multisig_guardian_ecdsa`, using `openzeppelin::auth::multisig_ecdsa` / `openzeppelin::auth::guardian_ecdsa`). That is a different implementation in a different namespace, so its procedure MAST roots don't match any standard template — the component is bucketed as `Custom`. `Custom` is not an auth component, so `interface.auth()` is empty and `account.getPublicKeyCommitments()` returns `[]`.

(Note: miden-standards *does* ship an `AuthGuardedMultisig` template, but it's a different implementation — `miden::standards::auth::*`, different exported procedures — so its roots can't match the OZ component.)

### Reproduction

1. Create a Guardian account via `MultisigClient.create(...)`.
2. Fetch it and call `account.getPublicKeyCommitments()` → `[]`.

A plain `AuthSingleSig` (BasicWallet) account returns its commitment as expected, so this is specific to the OZ component.

### Impact

Anything that reads a Guardian account's public key via the SDK breaks. In the Miden wallet this surfaced as **dApp connections failing with `NOT_GRANTED`** (the connect flow derives a session public key from `getPublicKeyCommitments()`), plus the reveal-private-key and public-key display views. Third-party dApps / SDK users reading a Guardian account's key hit the same.

### Request

Either:
[...]

PR #391, opened 2026-08-12
feat(multisig-client): expose signer public-key commitment accessors

haseebrabbani:
Stacked on #340 — merge only after it. Addresses #306.

## Context

#306 reported that `Account.getPublicKeyCommitments()` returned `[]` for Guardian accounts because miden-standards' `AccountInterface` did not recognize the previous OpenZeppelin auth component. With #340's move to the upstream `AuthGuardedMultisig` component, that call now works natively — this PR no longer "fixes" #306's symptom, and instead adds the Guardian-owned inspection API discussed in review: strict, layout-insulated accessors so consumers never read storage layout directly.

## What this adds

- `AccountInspector.getSignerPublicKeyCommitments(account)` — ordered approver commitments, validating the complete set against the configured signer count. Throws instead of silently omitting unreadable entries (upstream public-key extraction can omit them silently).
- `AccountInspector.getGuardianPublicKeyCommitment(account)` — the guardian commitment; throws when the entry is missing (the guarded-multisig always includes a guardian — no selector, no null case).
- `Multisig.getSignerPublicKeyCommitments()` / `getGuardianPublicKeyCommitment()` — instance variants reading current store-backed state.
- Both strict accessors are gated on #340's pinned-contract-version check (shared helper with `fromAccount`), so a foreign contract version is a clear error rather than misread storage.
- Fail-closed mutation path: `load()` and config refresh now validate the lenient `fromAccount()` result (complete signer set + guardian present) before adopting it, so a corrupt account can't be loaded and later rewritten without the omitted keys. `fromAccount()` stays lenient for inspect-only use.

## Robustness details

- Empty-word map reads are treated as "no entry" (`StorageMap::get` is `unwrap_or_default()` in miden-protocol; the contract zeroes removed entries) — never returned as `0x000…0` commitments.
- The untrusted signer count is bounded by a sanity ceiling (guards the ≥2^53 unsafe-integer loop case).
- A cross-SDK-copy `Word` rejection is translated into a descriptive error; the same-SDK requirement is documented.
- Slot names are shared between the storage writer and the readers via one internal module; deliberately not exported — consumers should use the accessors, not the layout.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -310,6 +310,17 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-cargo-npm-multisig-
 
+      # rust-toolchain.toml pins a toolchain with components that rustup syncs
+      # on first cargo use. Two test files shell out to `cargo run`
+      # concurrently (procedure-roots, account-roundtrip), and concurrent
+      # first-use syncs race each other ("detected conflict: 'bin/cargo'").
+      # Building both example fixtures here syncs once, serially, and turns
+      # the in-test cargo runs into cache hits.
+      - name: Sync pinned toolchain and prebuild example fixtures
+        if: matrix.package == 'miden-multisig-client'
+        working-directory: .
+        run: cargo build --quiet --example procedure_roots --example serialized_account -p miden-multisig-client
+
       - name: Install dependencies
         run: npm ci
 
--- a/crates/miden-multisig-client/examples/serialized_account.rs
+++ b/crates/miden-multisig-client/examples/serialized_account.rs
@@ -1,0 +1,68 @@
+//! Emit a deterministic serialized guarded-multisig account for the TS SDK's
+//! round-trip tests (`packages/miden-multisig-client/tests/account-roundtrip.test.ts`).
+//!
+//! The TS test deserializes the account with the real WASM SDK and asserts the
+//! `AccountInspector` accessors return the exact configuration written here,
+//! proving the Rust writer and TS reader agree on the storage layout and that
+//! the readers work against real SDK storage semantics.
+//!
+//! Run with:
+//! ```sh
+//! cargo run --example serialized_account -p miden-multisig-client
+//! ```
+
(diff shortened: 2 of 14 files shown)


zeljkoX (changes requested):
Thanks for implementing this. The strict accessors and real SDK round-trip coverage are thoughtful, and the implementation looks correct for the current local OpenZeppelin contract layout.

However, I do not think this should merge in its current form because #340 changes the underlying assumptions:

#340 replaces the local auth component with upstream AuthGuardedMultisig.

AccountInterface recognizes that component, so Account.getPublicKeyCommitments() should return the approver commitments natively. This resolves the original failure reported in #306.

The storage layout changes from openzeppelin::* to miden::standards::auth::*.
The guardian selector disappears because the upstream guarded-multisig component always includes a guardian.

As written, these accessors will therefore not work after #340. The signer accessor will read the old slots, while the guardian accessor will return null because it expects the removed selector.

I still see value in retaining a Guardian-owned inspection API:

- It provides strict complete-set validation, whereas upstream public-key extraction can silently omit unreadable entries.
- It insulates consumers from storage-layout changes.
- A focused guardian-commitment accessor is useful, even though the value is already available through AccountInspector.fromAccount().


Could we rebase this on #340 and adjust it to the upstream component?

Specifically:

- Use the miden::standards::auth::* storage layout.
- Remove guardian selector handling.
- Apply #340’s pinned-contract-version check to the strict accessors.
- Update the tests and documentation so the signer accessor is described as a strict and stable API rather than the fix required for getPublicKeyCommitments().
- Add a regression test confirming that account.getPublicKeyCommitments() works directly for an upstream AuthGuardedMultisig account.

haseebrabbani:
@zeljkoX Done — the PR is now stacked on #340: base retargeted to `upstream-contracts-016alpha` and the branch rewritten on top of its head (the previous 0.15-layout history is superseded). All five points addressed

zeljkoX (approved):
LGTM

Thanks for working on it!

Issue #387, opened 2026-08-11
multisig-client: move `nonce` into the options object of proposal-creation APIs


haseebrabbani:
## Context

`createP2idProposal(recipient, faucet, amount, nonce?, options?)` places the optional
`nonce` positionally before the options object, so callers who only want options must
pass a positional `undefined`:

```ts
await multisig.createP2idProposal(
  recipientAccountId,
  faucetAccountId,
  1000n,
  undefined,                    // nonce hole
  { reclaimHeight: 500_000 },
);
```

Raised during review of #381 ([comment](https://github.com/OpenZeppelin/guardian/pull/381#pullrequestreview-4896212440)):
the `undefined` hole makes every documented example ugly and the call sites error-prone.
The same shape applies to the rest of the `create*Proposal` family — options added by
future features (as #322 added `noteType` and #366 added `reclaimHeight`/`timelockHeight`)
make the positional `nonce` progressively worse.

## Proposal

Fold `nonce` into the options object so the 4th parameter is a single options bag:

```ts
await multisig.createP2idProposal(recipientAccountId, faucetAccountId, 1000n, {
  nonce: 42,                    // optional, defaults to Date.now() as today
  noteType: NoteType.Private,
  reclaimHeight: 500_000,
});
```

Two migration options:

1. **Deprecation window** — detect the 4th parameter's type (`number` ⇒ legacy nonce,
   object ⇒ options bag) and keep both working for a release, with the positional form
   marked `@deprecated`.
2. **Clean break** — change the signature outright in the next breaking release. If the
   #366 release ships as 0.17.0 (breaking on the Rust side anyway), this change could
   ride the same window and skip the type-detection dance entirely.

The wire format is unaffected either way — this is purely call-signature ergonomics.

## Scope

- `createP2idProposal`, `createConsumeNotesProposal`, `createAddSignerProposal`,
  `createRemoveSignerProposal`, `createChangeThresholdProposal`,
  `createSwitchGuardianProposal`, `createUpdateProcedureThresholdProposal`
- Example wrappers: `examples/_shared/multisig-browser`, `examples/web`, and the
  smoke-web harness
- Docs: `docs/MULTISIG_SDK.md`, `packages/miden-multisig-client/README.md`

PR #400, opened 2026-08-13
refactor(multisig-client): move nonce into the options object of proposal-creation APIs

haseebrabbani:
Closes #387. Based on #381 (`366-p2ide-heights`), riding the same breaking window.

Folds the positional optional `nonce` — and `newThreshold` for add/remove signer — into a single trailing options object across all eight `create*Proposal` methods (including `createCustomProposal`, so the family shares one convention):

```ts
await multisig.createP2idProposal(recipientAccountId, faucetAccountId, 1000n, {
  nonce: 42,                // optional, defaults to Date.now() as before
  noteType: NoteType.Private,
  reclaimHeight: 500_000,
});

### DIFF
--- a/docs/MULTISIG_SDK.md
+++ b/docs/MULTISIG_SDK.md
@@ -539,6 +539,12 @@ one implicitly.
 
 ### Proposal Operations
 
+Every `create*Proposal` method takes a single trailing options object (issue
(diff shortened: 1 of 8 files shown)


zeljkoX (approved):
Thanks for executing this one. 
LGTM
Breaking change makes sense given we would already have breaking change in the PR base branch.

PR #405, opened 2026-08-14
fix(multisig-client): prune executed proposals from the cache on sync

### LINKED ISSUE #404: syncProposals() keeps returning executed proposals as pending (stale cache never pruned)
### Summary

`Multisig.syncProposals()` in `@openzeppelin/miden-multisig-client` merges the GUARDIAN response into a persistent `this.proposals` Map but never prunes it. Once a proposal is executed, the GUARDIAN server drops it from `getDeltaProposals` (it only returns pending proposals), so a co-signer who did not execute the transaction locally keeps seeing the executed proposal as **pending — indefinitely**.

### Reproduction (2-of-2 multisig, two browsers)

1. Signer 2 creates a proposal and signs it (1/2).
2. Signer 1 signs it (2/2) and executes it.
3. Signer 2 syncs.

**Observed:** the executed transaction is still listed as a pending proposal on Signer 2, stuck at 1/2.
**Expected:** it disappears from Signer 2's pending list.

### Root cause

`syncProposals()` iterates the GUARDIAN deltas doing `this.proposals.set(...)`, then returns `Array.from(this.proposals.values())`. When GUARDIAN returns an empty (or reduced) list, the loop leaves previously-cached proposals in the Map. The signer who executed the tx updated its own cache locally as part of executing; every other signer's cache silently ignores the now-empty response and keeps the executed proposal pinned as pending.

### Versions

`@openzeppelin/miden-multisig-client`, `@openzeppelin/guardian-client`, and guardian server all on 0.16.x.

WiktorStarczewski:
### What

`syncProposals()` now reconciles the local proposal cache against the (authoritative) GUARDIAN response on every sync — proposals GUARDIAN previously reported and now omits are pruned — instead of only ever adding to a Map that was never cleaned up.

### Why

The Map was merge-only. After a proposal is executed, GUARDIAN stops returning it (the server keeps only pending proposals). The executing co-signer's browser had updated its own cache while executing, but every other signer's browser silently ignored the now-empty response and kept the executed proposal pinned as pending forever (stuck at, e.g., 1/2).

### How

- Track the ids GUARDIAN returned on the most recent sync (`lastReportedProposalIds`) and prune only those it now omits. A proposal GUARDIAN has never reported to this client — freshly created (before GUARDIAN's read-your-writes catches up) or imported-but-not-yet-synced — is never a prune candidate, so those flows keep working; a proposal GUARDIAN reported and then dropped (executed / abandoned) is pruned.
- Deliberately **no client-side nonce filter**. The proposal `nonce` field has no single cross-app convention (this repo's `examples/web` stores the pre-execution account nonce; `examples/_shared/multisig-browser` and the Rust client use the next nonce), so filtering on it in the shared client would evict live proposals for one of them. Nonce-based staleness *hiding* stays where it already lives — the callers' `filterVisibleProposals`.

### Tests

Three new `syncProposals` tests (each mutation-checked to fail against the bug it guards):
- **co-signer prune** — a proposal GUARDIAN reported then dropped is removed from the cache;
- **over-prune guard** — a proposal GUARDIAN still reports survives the reconcile;
- **read-your-writes** — a just-created proposal GUARDIAN has not reported yet is not evicted.

Full package suite green (426 tests), `tsc --noEmit` clean.

### Release / semver note

This changes the observable output of `syncProposals()` / `listProposals()` (they no longer retain proposals GUARDIAN has dropped). It fixes buggy behavior, and the in-repo example consumers already re-derive their visible set each sync, so no consumer regression is expected — but the release carrying it should flag this (a minor bump, or an explicit release note). No version bump is included here since versioning is a separate release commit.

### Notes for maintainers (out of scope for this PR)

- `examples/web/src/lib/multisigApi.ts` uses a different proposal-nonce convention (`proposalNonce = currentAccountNonce`, filter `nonce < accountNonce`) than `examples/_shared/multisig-browser` and the Rust client (`+1`, `nonce <= accountNonce`). Worth reconciling separately.
- If an offline-*create* path (a proposal not pushed to GUARDIAN at creation) is ever added, the prune would need to exempt session-local proposals; today every proposal is pushed at creation, so this can't arise.
- Optional follow-up (not needed for this fix): dedupe in-flight `syncProposals` calls (return the pending promise if one is running) to avoid a transient prune/keep blip when overlapping syncs race, and to cut redundant round-trips.

Closes #404

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved proposal synchronization to remove cached proposals no longer reported by GUARDIAN.
  * Preserved newly created or imported proposals until they are first reported by GUARDIAN.
  * Ensured currently reported proposals remain available in the cache.

* **Documentation**
  * Clarified proposal-cache cleanup behavior during synchronization.

### DIFF
--- a/packages/miden-multisig-client/README.md
+++ b/packages/miden-multisig-client/README.md
@@ -200,7 +200,7 @@ console.log('Signatures:', signedProposal.signatures.length);
 
 ### Sync Proposals
 
-Fetches proposals from the GUARDIAN server and updates local state:
+Fetches proposals from the GUARDIAN server and reconciles local state — proposals GUARDIAN no longer reports (executed, canonicalized, or abandoned) are pruned from the cache:
 
 ```typescript
 const proposals = await multisig.syncProposals();
--- a/packages/miden-multisig-client/src/multisig.test.ts
+++ b/packages/miden-multisig-client/src/multisig.test.ts
@@ -955,6 +955,121 @@ describe('Multisig', () => {
       expect(proposals[0].status).toBe('pending');
     });
 
+    it('should prune proposals GUARDIAN no longer reports (executed/canonicalized)', async () => {
+      const config = {
+        threshold: 2,
+        signerCommitments: ['0x' + 'a'.repeat(64), '0x' + 'b'.repeat(64)],
+        guardianCommitment: '0x' + 'c'.repeat(64),
+      };
+
+      const multisig = createTestMultisig(config);
+
+      const pendingProposal = {
+        account_id: '0x' + 'a'.repeat(30),
+        nonce: 1,
+        prev_commitment: '0x' + 'b'.repeat(64),
+        delta_payload: {
+          tx_summary: { data: 'AQID' },
+          signatures: [],
+          metadata: {
+            proposal_type: 'add_signer',
+            target_threshold: 1,
+            signer_commitments: ['0x' + 'a'.repeat(64)],
+            description: '',
+          },
+        },
+        status: {
+          status: 'pending',
+          timestamp: '2024-01-01T00:00:00Z',
+          proposer_id: '0x' + 'c'.repeat(64),
+          cosigner_sigs: [
+            {
+              signer_id: '0x' + 'a'.repeat(64),
+              signature: { scheme: 'falcon', signature: '0x' + 'e'.repeat(128) },
+              timestamp: '2024-01-01T00:00:00Z',
+            },
+          ],
+        },
+      };
+
+      // First sync: GUARDIAN reports the pending proposal, priming the cache.
+      mockFetch.mockResolvedValueOnce({
+        ok: true,
+        json: async () => ({ proposals: [pendingProposal] }),
+      });
+      const first = await multisig.syncProposals();
+      expect(first.length).toBe(1);
+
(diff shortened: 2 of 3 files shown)


zeljkoX (commented):
Thanks for this. Pruning only ids GUARDIAN previously reported is the right invariant, and TS-only is the right scope — Rust `list_proposals` is fetch-fresh.

Two things to fix before merge:

**1. Verify failure mid-loop leaves an unprunable proposal.** `verifyProposalMetadataBinding` is awaited inside the loop, so [P_ok, C_malformed] caches P and throws before `lastReportedProposalIds` is updated. P is then “never reported” and no later empty sync can prune it, the same forever pending shape as #404. Verify into a temp structure; merge and prune only once the whole response validates. 
**2. Overlapping syncs can delete a live proposal**. The prune loop reads `lastReportedProposalIds` at apply time(multisig.ts:597). Sync A fetches `[P]` and stalls in verify, `createProposal(Q)` caches Q, sync B fetches `[P, Q]` and completes, then A applies {P} and deletes Q. Prune turned a harmless overlap into a destructive one, so in-flight dedupe belongs in this PR.

zeljkoX:
Closing this one due to https://github.com/OpenZeppelin/guardian/pull/458 being merged.

PR #406, opened 2026-08-17
feat(auth): scope replay protection state per signer + retry

### LINKED ISSUE #367: Replay-protection CAS rejections surface as terminal "session expired" errors; classify them and align client timestamp/retry behavior
## Summary

Guardian's replay protection stores a strictly-increasing per-account auth timestamp and rejects any authenticated request whose `x-timestamp` doesn't win a compare-and-swap (CAS) against it. Legitimate concurrent requests can lose that race and receive:

```json
{
  "code": "authentication_failed",
  "message": "Your session has expired. Please sign in again.",
  "meta": { "retryable": false }
}
```

**Confirmed in production logs**: a user hit this while signing a proposal (note consumption) on a fresh account, and server logs show repeated `"Replay attack detected: timestamp not greater than last seen (CAS failed)"` warnings for the affected traffic.

Three distinct failure causes — CAS replay rejection, clock-skew rejection, and genuine signature failure — all collapse to this byte-identical wire response. Clients and support staff cannot distinguish them without server log access, and the two clients recover differently.

## Current behavior

For every authenticated request, `resolve_account` (`crates/server/src/services/mod.rs`):

1. Rejects timestamps more than 5 minutes from server time (`MAX_TIMESTAMP_SKEW_MS = 300_000`, `crates/server/src/metadata/auth/credentials.rs`) → `authentication_failed`.
2. Verifies the signature over account ID, timestamp, and request payload → `authentication_failed`.
3. Runs a CAS: `new_timestamp > last_auth_timestamp`, scoped **per account**. CAS failure → `authentication_failed`.

All three map to HTTP 401 / gRPC `Unauthenticated`, user message "Your session has expired. Please sign in again.", and `meta.retryable: false` (`crates/server/src/error.rs`). The distinguishing detail is deliberately log-only (feature 009 sanitized "Replay attack" off the wire).

Client behavior diverges:

- **TypeScript** (`packages/guardian-client/src/http.ts`): monotonic per-instance timestamps (`max(Date.now(), last + 1)`); retries `authentication_failed` up to 2 times with a fresh timestamp, fresh signature, and 50ms backoff. The retry branches on *every* `authentication_failed`, including terminal ones (invalid signature, skew), because it cannot tell them apart.
- **Rust** (`crates/client/src/client.rs`): raw `Utc::now().timestamp_millis()`, no monotonic guard, no retry. Additionally, the gRPC client currently drops the structured error code when mapping tonic statuses, so even today's `authentication_failed` is not machine-readable Rust-side.

## Why users see this despite the TS retry

A transient same-millisecond CAS race is healed by the TS retry and never surfaces. A user-visible failure means all three attempts lost within ~100ms — i.e. a **sustained** condition:

- **Clock-offset domination**: another client instance (cosigner device, second tab, background poller) with a clock ahead of the victim's — anywhere within the 5-minute skew window — recently authenticated. The stored timestamp then leads the victim's wall clock by more than the retry budget, so every fresh timestamp still loses.
- **In-flight reordering under continuous traffic**: monotonic generation prevents duplicate timestamps but not reordering — request T1 < T2 sent concurrently, T2 lands first, T1's CAS fails. Under polling (e.g. a background state refresh racing a proposal-signing request) this can repeat.
- Note: same-millisecond collisions from two instances only conflict transiently; sustained multi-instance failure specifically requires a clock offset between them.

Per-account scope makes this worse for multisig: all authorized cosigners across all devices compete on one timestamp, so one cosigner's fast clock locks out the others.

## Problems

1. CAS replay, clock skew, invalid signature, and unauthorized-key failures are wire-indistinguishable; clients cannot branch safely and support cannot diagnose without server logs.
2. Rust and TypeScript have divergent timestamp generation (monotonic vs raw) and recovery (retry vs none).
3. The TS retry fires on terminal failures too — harmless today but wrong once the causes are distinguishable.
4. `meta.retryable: false` contradicts actual TS client behavior for the replay case.
5. "Your session has expired. Please sign in again." is inaccurate for per-request signed auth — there is no session, and re-signing-in does not fix a clock offset or cosigner contention.
6. Per-account replay scope creates contention between independent authorized signers.

### 1. Stable replay-failure classification

Let clients distinguish a retryable timestamp replay from terminal authentication failures, e.g.:

```json
{
  "code": "authentication_failed",
  "meta": { "reason": "timestamp_replay", "retryable": true }
}
```

or a dedicated code such as `authentication_replay`. Clock skew, invalid signature, unauthorized signer, and malformed credentials stay non-retryable. Clients branch only on the stable code/typed metadata, never message text.

**Design note**: this partially reverses feature 009's decision to keep auth diagnostics off the wire. The proposal is that a coarse replay/terminal distinction is safe to expose — it only tells a replaying attacker their replay was detected, which is not a useful oracle — while signature-level detail stays log-only. This needs explicit sign-off.

**Process note**: any new code or meta field is a wire-contract change — it must follow the mandatory contract-change workflow (`AGENTS.md` §4), be added to `GUARDIAN_ERROR_CODES`/`WIRE_ALIASES` in `packages/guardian-client/src/error-codes.ts`, and regenerate committed OpenAPI specs.

### 2. Align timestamp generation

Both clients guarantee strictly increasing timestamps per instance:

```
next_timestamp = max(current_time_ms, previous_timestamp + 1)
```

TS already does this; port to the Rust client. Time source injectable for tests.

### 3. Align bounded retry behavior

Both clients: retry **only** an explicitly classified replay rejection (per change 1); fresh timestamp, recomputed digest, fresh signature, identical payload each attempt; same bounded count and backoff; never retry terminal failures. Prerequisite for Rust: plumb the structured error code through the gRPC client's status mapping so it is machine-readable at all.

### 4. Re-evaluate replay-protection scope

Decide whether replay state stays per-account or moves to per-(account, signer commitment). Per-signer scope removes cross-cosigner and cross-device contention while still rejecting replays from the same signer — it directly addresses the multisig lockout above. Document the chosen security model and concurrency semantics.

**Sequencing**: #365 (`account_auth_state` split) moves replay state into a dedicated narrow table/file and removes `AccountMetadata.last_auth_timestamp`. This change must land after #365 and be designed against the new store; per-signer scope becomes a key-shape change to `account_auth_state`.

### 5. Fix user-facing messaging

Replace "Your session has expired. Please sign in again." for per-account signed requests with wording that matches the auth model, e.g.:

> Guardian could not authenticate this request. Please retry or reconnect your signer.

Dashboard and EVM session-auth surfaces may retain session-specific wording where an actual session exists.

## Acceptance criteria

- Replay/CAS rejection is machine-classifiable on both HTTP and gRPC without parsing text, with consistent `retryable` metadata.
- Rust and TS use equivalent monotonic timestamp generation and equivalent bounded replay-only retry (fresh timestamp + signature per attempt).
- Terminal authentication failures (invalid signature, unauthorized signer, malformed credentials, outside-skew timestamps) are never retried.
- Rust gRPC client surfaces structured error codes.
- Tests cover: same-millisecond requests from one instance; concurrent requests from multiple instances; concurrent requests from different authorized cosigners; clock-ahead-within-skew domination; outside-skew rejection; in-flight reordering under concurrency; both Falcon and ECDSA auth paths.
- OpenAPI specs regenerated if the HTTP error contract changes; contract-change workflow followed.
- `docs/TROUBLESHOOTING.md` gains a symptom entry for this 401 (currently the top support-confusing error) describing the replay model, and auth docs/spec diagrams (`spec/processes.md`) reflect the final design.

## Impacted areas

- `crates/server/src/services/mod.rs` (`resolve_account`)
- `crates/server/src/metadata/` — replay state store (post-#365: `account_auth_state` in `postgres.rs` / `auth_state.json` in `filesystem.rs`)
- `crates/server/src/error.rs`, `crates/server/src/api/http.rs`, `crates/server/src/api/grpc.rs`
- `crates/client/src/client.rs`, `crates/client/src/error.rs`
- `packages/guardian-client/src/http.ts`, `packages/guardian-client/src/error-codes.ts`
- Server, Rust-client, and TS-client tests; `docs/TROUBLESHOOTING.md`; `spec/processes.md`; committed OpenAPI specs

## Validation

```
cargo test -p guardian-server
cargo test -p guardian-client
cd packages/guardian-client && npm test
```

If behavior reaches the multisig proposal workflow:

```
cargo test -p miden-multisig-client
cd packages/miden-multisig-client && npm test
```

zeljkoX:
Scope replay protection per account and signer.
Add retryable authentication_replay error handling and bounded SDK retries.
Safely migrate PostgreSQL and filesystem replay state.

## Compatibility and rollout notes

Successful `POST /configure` responses keep their existing shape. Non-2xx HTTP
responses now use Guardian's standard `{ code, message, meta }` error envelope
instead of a `ConfigureResponse` with `success: false`. Direct HTTP consumers
that inspect `success` on configure errors must switch to the HTTP status and
stable error `code`. The published TypeScript client already throws
`GuardianHttpError` for non-2xx responses.

Clients retry only `authentication_replay`; they never retry
`authentication_failed`. During a mixed server/client rollout, a replay CAS
reported under the older authentication code—or received by a client without
replay-specific retry handling—can therefore surface as a terminal 401 until
both sides use the same error contract. This PR documents that rollout window
instead of adding a compatibility shim.

The filesystem startup migration validates the complete in-memory expansion
before rewriting `auth_state.json`. Filesystem stores from an earlier revision
without a reserved account floor are forward-repaired from the maximum existing
signer timestamp. No PostgreSQL repair migration is needed because no database
ran an earlier revision of this PR.

First-time `/configure` must create account metadata before seeding replay state
because the replay row has a foreign key to the account. A crash or storage
failure in that narrow interval can leave the new account without a floor, so a
captured request that is still inside the five-minute skew window can initialize
the floor and execute. Replaying configure itself is an idempotent re-submit;
the accepted bounded window is documented in `spec/api.md`.

## Summary by CodeRabbit

- **New Features**
  - Added automatic replay-protection retries for signed requests with refreshed timestamps and signatures.
  - Supports independent replay protection for multiple authorized signers on the same account.
  - Added the `authentication_replay` error code for retryable authentication conflicts.

- **Bug Fixes**
  - Terminal authentication failures are no longer retried unnecessarily.
  - Improved handling of concurrent requests, clock changes, and request reordering.
  - Standardized configuration error responses.

- **Documentation**
  - Updated client, API, migration, scaling, and troubleshooting guidance for replay protection and authentication errors.

### DIFF
--- a/crates/client/Cargo.toml
+++ b/crates/client/Cargo.toml
@@ -18,7 +18,7 @@ serde_json = { workspace = true }
 tonic = { workspace = true }
 tonic-prost = { workspace = true }
 prost = { workspace = true }
-tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
+tokio = { workspace = true, features = ["macros", "rt-multi-thread", "time"] }
 thiserror = { workspace = true }
 hex = { workspace = true }
 chrono = { workspace = true }
--- a/crates/client/README.md
+++ b/crates/client/README.md
@@ -55,8 +55,8 @@ class), and `retry_after()` returns the server's hint, preferring the
 `retry-after` status metadata over the envelope value.
 
 Rate-limit rejections happen before the server touches any state, so
-retrying them is always safe. The client does not retry automatically
-(automatic backoff is tracked in
+retrying them is always safe. The client does not retry rate limits
+automatically (automatic backoff is tracked in
 [#360](https://github.com/OpenZeppelin/guardian/issues/360)); a bounded
 loop over the exposed hint is a few lines:
 
@@ -79,6 +79,18 @@ let state = loop {
 Retries are idempotent (the original request timestamp is preserved). The
 server refuses with `GUARDIAN_CANDIDATE_LANDED` when the transaction
 demonstrably landed.
+
+### Replay-protection retries
+
+Signed requests carry a strictly increasing per-instance timestamp
+(`max(now_ms, previous + 1)`). When a correctly signed request still loses
+the server's per-signer replay check (stable code `authentication_replay`,
+typically two in-flight requests landing out of order), the client retries
+automatically, up to 2 times with a 50ms backoff, minting a fresh timestamp
+and signature over the identical payload each attempt.
+`ClientError::is_replay_rejection()` exposes the classification. Terminal
+authentication failures (`authentication_failed`: clock outside the skew
+window, invalid or unauthorized signature) are never retried.
 
 ## Authentication
 
--- a/crates/client/README.md
+++ b/crates/client/README.md
@@ -92,6 +92,12 @@ and signature over the identical payload each attempt.
 authentication failures (`authentication_failed`: clock outside the skew
 window, invalid or unauthorized signature) are never retried.
 
+The client retries only `authentication_replay`; it never retries
+`authentication_failed`. During a mixed server/client rollout, a replay CAS
+reported under the older authentication code—or received by a client without
+replay-specific retry handling—can therefore surface as a terminal error until
+both sides use the same error contract.
+
 ## Authentication
 
 The client uses Falcon Poseidon2 signatures for authenticated requests. Here is how to set it up:
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -509,12 +509,19 @@ impl GuardianClient {
 /// instance; the caller supplies the clock so the arithmetic stays
 /// deterministic under test.
 fn next_monotonic_timestamp(last_timestamp: &AtomicI64, now_ms: i64) -> i64 {
-    let previous = last_timestamp
-        .fetch_update(Ordering::SeqCst, Ordering::SeqCst, |last| {
-            Some(last.max(now_ms - 1) + 1)
-        })
-        .expect("update closure always returns Some");
-    previous.max(now_ms - 1) + 1
+    let mut previous = last_timestamp.load(Ordering::SeqCst);
+    loop {
+        let next = previous.saturating_add(1).max(now_ms);
+        match last_timestamp.compare_exchange_weak(
+            previous,
+            next,
+            Ordering::SeqCst,
+            Ordering::SeqCst,
+        ) {
+            Ok(_) => return next,
+            Err(actual) => previous = actual,
+        }
+    }
 }
 
 fn attach_auth_headers<T: prost::Message>(
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -21,8 +21,18 @@ use guardian_shared::lookup_auth_message::LookupAuthMessage;
 use miden_protocol::Word;
 use miden_protocol::account::AccountId;
 use std::sync::Arc;
+use std::sync::atomic::{AtomicI64, Ordering};
+use std::time::Duration;
 use tonic::metadata::MetadataValue;
 use tonic::transport::Channel;
+
+/// Bounded retry policy for replay-classified rejections
+/// (`authentication_replay`): the request was correctly signed but lost the
+/// server's per-signer timestamp CAS to a concurrent request. Each attempt
+/// re-mints the timestamp and re-signs the identical payload. Mirrors the TS
+/// client's policy; terminal authentication failures are never retried.
+const REPLAY_RETRY_LIMIT: u32 = 2;
+const REPLAY_RETRY_BACKOFF: Duration = Duration::from_millis(50);
 
 /// A client for interacting with Guardian servers.
 ///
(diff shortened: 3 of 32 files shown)


haseebrabbani (approved):
lgtm, minor comments:

1. **First-time configure residuals** (documented, just confirming they're accepted): a crash between `metadata.set` and the floor seed leaves a new account floorless, and two concurrent first-configures can surface `authentication_replay` for a configure that actually persisted (client retry heals via the reconfigure path).
2. **Floor-repair marker** (`auth_state_legacy_floor_v1`): if the marker is lost (restore from a backup predating it), the repair re-runs and synthesizes account floors from max signer timestamps for *all* accounts, including post-migration ones — a fresh signer with a trailing clock then loses its first CAS once. Self-limiting, but worth a line in the TROUBLESHOOTING restore section.
3. **Version skew** stays a docs-only mitigation — old clients won't retry `authentication_replay` during the rollout window. Fine if consciously accepted, which the docs now make explicit.

Also flagging for visibility: HTTP `/configure` failures switching to the standard error envelope is a breaking change for direct HTTP integrations (documented, but easy to miss in a replay-protection PR).

PR #408, opened 2026-08-17
chore(server): serve status content on GET / instead of "Hello, World!"

### LINKED ISSUE #403: Remove Hello, World! on GET / — serve status content instead
## Summary

`GET /` currently returns a hard-coded `"Hello, World!"`, while `GET /status` already provides unauthenticated, machine-readable health information.

## Proposed change
Replace the root handler with an alias to `GET /status`.

```
Router::new().route("/", get(status))
```

This keeps existing 200 OK health checks working while returning useful status data, including version, environment, start time, and uptime.
Document GET / in OpenAPI as a /status alias and regenerate the committed specifications.

## Acceptance criteria

- `GET /` returns the same JSON response as `GET /status`.
- The `"Hello, World!"` root handler is removed.
- `GET /` remains unauthenticated.
- Tests verify that `/` and `/status` return equivalent responses.
- OpenAPI specifications are updated.

zeljkoX:
Closes #403.

`GET /` returned a hard-coded `"Hello, World!"` while `GET /status` already served unauthenticated, machine-readable health information. Root probes now get the same status body: version, git commit, environment, start time, uptime.

## Changes

- `crates/server/src/api/http.rs` — new `status_root` handler delegating to `status`, so both routes return identical bodies by construction. It carries its own `#[utoipa::path(get, path = "/")]` annotation; routing `/` straight to `get(status)` would have left the root undocumented.
- `crates/server/src/builder/handle.rs` — `root()` and the `"Hello, World!"` literal removed; `.route("/", get(status_root))`.
- `crates/server/src/openapi.rs` — `status_root` registered on `ClientApiDoc`. The spec test now asserts `/` is present and carries no security requirement.
- `crates/server/src/testing/helpers.rs` — the test router mounts `/`, so the equivalence test exercises real axum routing rather than calling the handler directly.
- `docs/openapi.json`, `docs/openapi-client.json` — regenerated with `cargo run --features evm --bin gen-openapi -- docs`.
- `spec/api.md` — endpoint catalog gains `GET /` and `GET /status` (the latter was missing from the table entirely).
- `docs/QUICKSTART.md`, `docs/LOCAL_DEV.md` — the `curl http://localhost:3000/` liveness lines now say what comes back.

## Breaking change

`GET /` changes content type and body: `text/plain` `"Hello, World!"` becomes `application/json` `StatusResponse`. The status code is unchanged (200) and the route stays unauthenticated, so status-code-only health checks are unaffected.

Verified nothing consumes the old body:

- `infra/alb.tf` target group health-checks `path = "/"` with `matcher = "200"`, status code only.
- The Caddy health probe targets `/`, status code only.
- No code in `crates/`, `packages/`, or `examples/` requests the bare root, and no `Hello, World` string remains in the repo.

Anyone with an external monitor asserting on the response body will need to update it.

## Client parity (deliberate skip)

`packages/guardian-client` already exposes `getStatus()` against `/status` (`src/http.ts:243`, tested at `src/http.test.ts:135`). No client change is needed, and adding a second method pointing at `/` would be a redundant path to identical data, so none was added.

Noting a pre-existing asymmetry found while checking this: `crates/client` has no status method at all, while the TypeScript client does. Out of scope here.

## Testing

- `root_route_returns_the_status_body_without_auth` (`crates/server/src/api/http.rs`) asserts both routes return 200 unauthenticated with byte-identical JSON. `uptime_seconds` is asserted to be a number rather than compared for equality: it comes off the system clock, so two back-to-back requests can straddle a second boundary. Every other field is compared exactly.
- `openapi_spec_builds_and_serializes` asserts `/` is documented and public.
- `cargo test -p guardian-server --lib` — 855 passed, 0 failed.
- `cargo clippy -p guardian-server --all-targets --features evm` — clean.
- `cargo fmt --all -- --check` — clean.
- `cargo run --features evm --bin gen-openapi -- --check docs` — specs in sync (the `OpenAPI Spec Drift` CI job).

## Summary by CodeRabbit

* **New Features**
  * Added a public `GET /` endpoint that returns the same status information as `GET /status`.
  * Documented the endpoint in the API catalog and OpenAPI specifications.
* **Documentation**
  * Updated local development and quickstart guides with the new liveness-check alias.
* **Tests**
  * Added coverage confirming both endpoints return equivalent successful responses.
  * Verified the endpoint is publicly accessible in the API specification.

### DIFF
--- a/crates/server/src/api/http.rs
+++ b/crates/server/src/api/http.rs
@@ -735,34 +735,34 @@ mod tests {
         (state, storage, network, metadata)
     }
 
-    #[tokio::test]
-    async fn status_route_returns_ok_without_auth() {
+    async fn get_json(state: AppState, uri: &str) -> serde_json::Value {
         use crate::testing::helpers::create_router;
         use axum::body::{Body, to_bytes};
         use axum::http::{Request, StatusCode};
         use tower::ServiceExt;
 
-        let (state, ..) = create_test_state();
-        let app = create_router(state);
-
-        let res = app
+        let res = create_router(state)
             .oneshot(
                 Request::builder()
                     .method("GET")
-                    .uri("/status")
+                    .uri(uri)
                     .body(Body::empty())
                     .unwrap(),
             )
             .await
-            .expect("status request should succeed");
-
-        assert_eq!(res.status(), StatusCode::OK);
+            .expect("request should succeed");
+        assert_eq!(res.status(), StatusCode::OK, "{uri} should return 200");
 
         let bytes = to_bytes(res.into_body(), usize::MAX)
             .await
-            .expect("status body should read");
-        let json: serde_json::Value =
-            serde_json::from_slice(&bytes).expect("status body should be JSON");
+            .expect("body should read");
+        serde_json::from_slice(&bytes).expect("body should be JSON")
+    }
+
+    #[tokio::test]
+    async fn status_route_returns_ok_without_auth() {
+        let (state, ..) = create_test_state();
+        let json = get_json(state, "/status").await;
 
         assert_eq!(json["status"], "ok");
         assert!(json["version"].is_string());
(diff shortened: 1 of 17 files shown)


haseebrabbani (approved):
lgtm, minor comments:

1. **The production `/` route is untested.** The new test uses `testing::helpers::create_router`, whose `/` route this PR added by hand — not the router in `handle.rs`. Dropping the route from `handle.rs:246` leaves the suite green while production `/` 404s and the ALB drains targets. Sharing one route table between `run()` and the test helper would fix it.
2. **`docs/OPENAPI.md`** still says public endpoints are "(`/pubkey`)" — there are now three (`/pubkey`, `/status`, `/`).
3. `get_json` duplicates the request/decode sequence open-coded in the sibling status test; hoist it to `mod tests` scope and reuse.
4. Nothing keeps the `/` and `/status` annotations in sync if `/status`'s contract changes — a `Modify` addon cloning the `/status` `PathItem` under `/` would express the alias once.

PR #410, opened 2026-08-17
fix(multisig-client): block-independent metadata↔tx_summary binding

### LINKED ISSUE #409: Second signer cannot load a multisig account while a proposal is pending (block-dependent metadata binding)
### Summary

A second signer cannot load a 2-of-2 multisig account while any proposal is pending. Load fails with:

```
Invalid proposal: metadata does not match tx_summary for 0x…
```

Once no proposals are pending, the second signer loads fine.

### Reproduction

1. Signer 1 creates a proposal and leaves it pending (does not execute).
2. Signer 2 (fresh browser, never loaded this account) tries to load the account → **fails completely**.
3. Execute/clear the pending proposal → Signer 2 loads successfully.

Setup: `@openzeppelin/miden-multisig-client@0.16.1`, `@openzeppelin/guardian-client@0.16.1`, `@miden-sdk/miden-sdk@0.15.9`, Miden testnet, 2-of-2.

### Root cause

On load, `syncProposals()` runs `verifyProposalMetadataBinding()` on every pending proposal. For built-in proposal types that rebuilds the transaction from the proposal's metadata and **re-executes it** (`executeForSummary`) to check the resulting `TransactionSummary` commitment equals the one the proposer signed.

That re-execution is **block-height dependent**:
- `executeForSummary` passes no reference block; `execute_transaction` uses the client's *current sync height* as the reference block (miden-client `transaction/mod.rs`: "The transaction will use the current sync height as the block reference").
- `TransactionSummary.to_commitment()` hashes `{account_delta, input_notes, output_notes, salt}`, and `account_delta` includes the **transaction fee**, which is computed from the reference block header's `fee_parameters` (miden-base `executor/mod.rs`).

So the same logical proposal re-executed at a different block yields a different summary commitment. The check only passes on the same client at the same sync height as the proposer — which is exactly the reporter's evidence: the proposer's re-verify at sync height 1549420 matches; a second signer at 1549447 gets a different commitment and the whole load aborts.

The `custom` and `switch_guardian` proposal types were already exempt from this re-execution; the bug is that every other type still runs it.

WiktorStarczewski:
> **Draft — depends on `@miden-sdk/miden-sdk@0.15.10` (companion PR 0xMiden/web-sdk#298) being published, and on the real-summary integration test below. Do not un-draft until both are done.**

### What

Replaces the block-height-dependent metadata re-execution in `verifyProposalMetadataBinding` with a **deterministic, exhaustive** metadata↔`tx_summary` binding. Fixes the cross-block liveness bug (a second signer couldn't load/sign a pending proposal) while **preserving** the documented what-you-see-is-what-you-sign guarantee (`docs/MULTISIG_SDK.md`, `docs/CONCEPTS.md`: "a compromised Guardian cannot steal funds").

### Why

`verifyProposalMetadataBinding` rebuilt each proposal from metadata and **re-executed** it (`executeForSummary`) to compare the resulting `TransactionSummary` commitment to the signed one. That re-execution runs against the client's **current sync height** as the reference block, and the summary commits over the **transaction fee**, derived from that block's `fee_parameters`. So the check only matched on the same client at the same height as the proposer — a second signer at a later block hit `metadata does not match tx_summary` and the account load aborted.

Simply *deleting* the check would fix liveness but drop a documented security control: a compromised/MITM Guardian could serve a cosigner an evil `tx_summary` under benign `metadata`, and the cosigner — reviewing metadata but signing the opaque summary commitment — could be tricked into signing a drain. So the check is made deterministic instead.

### How

The new `assertMetadataMatchesSummary` (`src/multisig/summaryBinding.ts`) decodes the signed summary and, per type, asserts the transaction's effects are **exactly** the declared effect + the fee, across **every** intent-bearing dimension — its output notes, its input notes, and the account-storage slots it changes. It never reads the vault delta (the fee is a block-dependent native-asset vault delta); that's safe because assets can only leave via an output note or the fee, so binding the notes + storage slots exactly makes the vault a consequence. The comparison is therefore deterministic across blocks yet still exact — a mislabel, or a piggybacked undeclared effect in any dimension, is rejected.

Per type: **p2id** — exactly one output note (recipient rebuilt from metadata + the summary's salt; exactly the declared fungible asset), no input notes, no storage change but the per-tx replay marker. **consume_notes** — input-note-id set equality, no output notes. **signer** — no notes; `threshold_config = [threshold, count, 0, 0]`; the `signer_public_keys` map delta index-bound to the declared set (catches a duplicated/omitted/injected signer). **update_procedure_threshold** — no notes; `procedure_thresholds[PROC_ROOT] = [threshold, 0, 0, 0]`. **switch_guardian** — now bound (was exempt): guardian pubkey map = `newGuardianPubkey`, storage confined to the guardian slots (+ signer slots if it also rotates signers). **custom** — exempt (no metadata recipe; WYSIWYS explicitly does not hold — cosigners must verify the raw `tx_summary`).

Reading the config-type map deltas needs `AccountStorageDelta.maps()` / `valueDeltas()`, added in `@miden-sdk/miden-sdk@0.15.10` (**0xMiden/web-sdk#298**); this PR bumps the dep to `^0.15.10`.

### Tests

35 binding unit tests including a **rejecting** test for each fund-theft bypass an earlier field-allowlist draft had (consume+drain-note, p2id piggybacked storage write / signer-map write / extra asset, signer duplicate-with-omit, unexpected slot, proc/guardian + drain note, wrong guardian key). Full package suite green (458), `tsc --noEmit` clean.

### Known gaps

- **⚠️ Non-fungible asset drain in p2id (fund theft) — closure known, must land before merge.** The p2id binding checks the recipient and the *fungible* asset separately, and `NoteAssets` exposes only fungible assets — so an NFT attached to an otherwise-correct p2id note is not bound and could be drained under a "send N tokens" label. Closure: bind the output note by full `outputNote.id()` equality (a `NoteId` commits to *all* assets incl. NFTs; the `consume` path already binds by note id and `p2id.ts` has the note-ID reconstruction), sourcing the fungible asset off the note to preserve block-independence. This also binds the note *type* as a side effect. It has callback-flag/metadata caveats that the pre-merge integration test must validate — so it's tracked below, not shipped in this draft.
- **`custom`** proposals get no WYSIWYS — cosigners must verify the raw `tx_summary` (unavoidable; no metadata recipe).
- **Signer unchanged-index**: the storage-map delta carries only *changed* indices, so an unchanged declared index isn't re-verified (a kept index can only hold the pre-existing legitimate signer; an omitted signer is under-delivery/DoS caught at execution, not theft).

### ⚠️ Pre-merge checklist (do not un-draft until all done)

- [ ] `@miden-sdk/miden-sdk@0.15.10` published (web-sdk#298 merged + released).
- [ ] **Close the p2id NFT gap** via `outputNote.id()` equality (see above).
- [ ] **Real-summary integration test.** Every unit test mocks the `TransactionSummary`, so the encodings are cross-verified by reading the builders/MASM/inspector but not against a genuine executed summary. Add a test that runs a real `executeForSummary → serialize → deserialize → assertMetadataMatchesSummary` per type and confirms: (a) **private p2id** passes (its own-output note must expose `assets()`, else it would false-reject); (b) the `threshold_config` / procedure-root / guardian-pubkey word encodings match a real delta; (c) the signer pubkey-map delta carries every changed index; (d) the NFT-gap closure doesn't false-reject an honest p2id (callback-flag / note-metadata handling).

### Follow-ups (separate)

- The sibling binding in `prepareCustomExecution` (custom execute path) has the same block-dependence; give it the equivalent treatment.
- The Rust reference client (`crates/miden-multisig-client`) still re-executes on list/sign — same bug, wants the equivalent block-independent binding.

Closes #409

### DIFF
--- a/docs/MULTISIG_SDK.md
+++ b/docs/MULTISIG_SDK.md
@@ -403,12 +403,18 @@ SDK's own built-in execution uses. Extending a transaction's advice map with it
 therefore does not collide with the transaction's ordinary inputs; the
 integration extends rather than replaces its advice map.
 
-> **Security:** for first-party types the SDK reconstructs the transaction from
-> metadata and checks it against the signed `tx_summary` commitment. For custom
-> types there is no such reconstruction, so the SDK cannot verify that display
-> metadata (e.g. `description`) matches what the transaction actually does.
-> Cosigners must verify the raw `tx_summary` they are signing — not trust the
-> label or description.
+> **Security:** for first-party types the SDK verifies that the signed
+> `tx_summary` matches the proposal's metadata before a cosigner signs it. It
+> decodes the summary and asserts the transaction's effects — its output notes,
+> its consumed input notes, and the account-storage slots it changes — are
+> *exactly* what the metadata describes and nothing more, so a cosigner is shown
+> what they actually sign. The check is deterministic (it never reads the
+> block-dependent transaction fee, which is why it works across cosigners at
+> different sync heights). Two residual gaps: it cannot see **non-fungible**
+> assets (the SDK exposes only fungible note/vault assets), and for **custom**
+> types there is no metadata recipe at all, so the SDK cannot verify that display
+> metadata (e.g. `description`) matches what the transaction does. Cosigners must
+> verify the raw `tx_summary` for `custom` proposals — not trust the label.
 
 ### Offline Workflow
 
@@ -584,13 +590,15 @@ await multisig.exportNoteToFile(noteId);
 const importedNoteId = await multisig.importNoteFromBytes(noteFileBytes);
 ```
 
-> **Note:** every cosigner device that verifies or signs the consume-notes
-> proposal needs the note in its local store with the on-chain inclusion
-> proof — deliver the note file to each of them (import + sync), not just to
-> the proposer. A cosigner whose store lacks the authenticated note rebuilds
-> the transaction differently (the input-notes commitment distinguishes
-> authenticated from unauthenticated consumption) and rejects the proposal
-> with `metadata does not match tx_summary`. The sender's own device heals
+> **Note:** the device that **executes** the consume-notes proposal needs the
+> note in its local store with the on-chain inclusion proof — deliver the note
+> file to each cosigner (import + sync), any of whom may execute, not just to
+> the proposer. Verifying and signing no longer require the note: the SDK checks
+> the proposal by comparing the signed summary's input-note **ids** to the
+> metadata (note ids are proof-agnostic), so a cosigner can sign without the
+> note. But the device that executes rebuilds the consumption from its own store,
+> so a store lacking the authenticated note produces a transaction that no longer
+> matches the signed summary and fails on-chain. The sender's own device heals
 > itself: it already knows the full note, so a post-commit sync is enough.
 
 #### Consume Notes (Claim Received Funds)
--- a/packages/miden-multisig-client/package.json
+++ b/packages/miden-multisig-client/package.json
@@ -36,7 +36,7 @@
     "test:watch": "npm run generate:masm && vitest"
   },
   "dependencies": {
-    "@miden-sdk/miden-sdk": "^0.15.8",
+    "@miden-sdk/miden-sdk": "^0.15.10",
     "@noble/curves": "^1.9.7",
     "@noble/hashes": "^2.0.1",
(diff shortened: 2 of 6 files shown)


zeljkoX:
@haseebrabbani Can you check, can we close this PR now given #409 is merged?

PR #411, opened 2026-08-17
chore: Add Postgres CI tests

### LINKED ISSUE #389: Run Postgres-backed server tests in PR CI
## Problem

The server currently has 16 Postgres-backed tests that never run in an automated pipeline:

| Module | Ignored tests |
| --- | ---: |
| `crates/server/src/metadata/postgres.rs` | 7 |
| `crates/server/src/storage/postgres.rs` | 5 |
| `crates/server/src/audit/postgres.rs` | 1 |
| `crates/server/src/coordination/postgres/lease.rs` | 1 |
| `crates/server/src/coordination/postgres/session_store.rs` | 1 |
| `crates/server/src/coordination/postgres/challenge_store.rs` | 1 |
| **Total** | **16** |

These tests are behind the `postgres` Cargo feature and marked `#[ignore]` because they require a live database through `DATABASE_URL`. Most use:

```rust
#[ignore = "requires DATABASE_URL with migrations applied"]
```

The migration-backfill test has a more specific ignore reason because it reverts and reapplies the newest migration.

CI compiles the Postgres implementation through:

```bash
cargo clippy --all-targets --all-features
cargo build --release --all-features
```

However, it never executes these tests. The coverage step uses default features, and ignored tests would remain skipped even in a full-feature test run.

Unlike the gaps tracked in #388, these suites do not run during the Rust crate publication workflow: `guardian-server` is not a published crate, and ignored tests are skipped unless explicitly requested. In practice, they run only when a developer manually provisions Postgres and remembers to opt into them.

The cost is concrete. The test:

```text
migration_backfills_legacy_timestamps_into_account_auth_state
```

was added in #368 after review identified a migration lock-discipline gap. It protects a schema migration path but currently runs zero times per PR and zero times per release.

The lease, session-store, and challenge-store tests also cover multi-instance coordination invariants that have no non-Postgres equivalent.

## Proposal

Add a dedicated Postgres-backed test job to PR CI.

1. Attach a Postgres service container with a health check so the test does not start before the database is ready.
2. Export `DATABASE_URL` pointing at the service database.
3. Use a fresh database. Each Postgres-backed test invokes the embedded migrations through `run_migrations()`, so no separate CI migration step is required. The migration-backfill test manages its own revert/reapply cycle.
4. Run only the ignored tests whose fully qualified names contain the `postgres` module path:

```yaml
- name: Run Postgres-backed tests
  env:
    DATABASE_URL: postgres://postgres:postgres@localhost:5432/guardian_test
  run: >
    cargo test -p guardian-server --lib --features postgres
    -- --ignored postgres --test-threads=1
```

The `postgres` name filter is necessary because bare `--ignored` would also execute unrelated ignored tests, including live-network checks and timing benchmarks.

The suites should initially run with `--test-threads=1`. Although most tests create uniquely named state, the migration-backfill test temporarily reverts the newest migration for the shared database. Its module-local lock does not serialize tests in the storage, audit, or coordination modules.

Running this as a dedicated job keeps the existing default-feature coverage path unchanged and makes Postgres-specific failures easier to identify. Database-per-test isolation or a crate-wide database lock could allow parallel execution later.

## Acceptance criteria

- A failing Postgres-backed ignored test in `guardian-server` fails PR CI.
- `migration_backfills_legacy_timestamps_into_account_auth_state` runs on every PR.
- The Postgres test command does not execute unrelated ignored tests.
- The existing default-feature coverage job remains unchanged.

## Related

- #368 added the migration-backfill regression guard.
- #388 tracks integration tests and doctests excluded by the coverage step’s `--lib` scope. Postgres-backed tests remain separate because their fix requires a service container, the `postgres` feature, and explicit ignored-test selection.

zeljkoX:
This PR improves CI workflow to run additional tests that required dependencies like postgres db.

## CI notes

The Postgres job newly makes two timing-sensitive regression tests required in CI:

- `pull_candidate_deltas_stays_flat_under_deep_history` requires the filtered read to be at least 5x faster.
- `lease_transfer_does_not_wait_for_a_validated_write_transaction` uses a 1,200 ms transaction hold.

If this job flakes under shared-runner load, check these tests first before treating the failure as a Postgres setup problem.

## Summary by CodeRabbit

* **New Features**
  * Added automated PostgreSQL-backed integration and end-to-end test coverage.
  * Added safeguards for test database setup, schema resets, migrations, and readiness checks.
  * Added a dedicated script for running PostgreSQL tests locally.

* **Documentation**
  * Documented PostgreSQL test setup, configuration, execution, and safety requirements.
  * Clarified how to run live network integration tests.

* **Tests**
  * Added HTTP body-limit integration test coverage.
  * Improved consistency and reliability of PostgreSQL test execution.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,6 +70,13 @@ jobs:
       - name: Install cargo-llvm-cov
         uses: taiki-e/install-action@4a44d353abec8bd780e47de38e5bc5ea77053391 # cargo-llvm-cov
 
+      # guardian-server/postgres is enabled so the ~50 tests behind that feature
+      # which need no database (TLS mode resolution, certificate verification,
+      # credential redaction, fail-closed store guards, the test-database name
+      # guard) actually execute. Clippy and Build compile them via --all-features
+      # but run nothing, and no other job enables the feature, so before this
+      # they ran in no job at all. The database-backed tests stay #[ignore] here
+      # and run in the postgres-test job, so this job needs no database.
       - name: Run tests with coverage
         run: >-
           cargo llvm-cov
@@ -78,6 +85,7 @@ jobs:
           --exclude guardian-server-bench-loadgen
           --exclude guardian-demo
           --exclude guardian-rust-example
+          --features guardian-server/postgres
           --lcov
           --output-path lcov.info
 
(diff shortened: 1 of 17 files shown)


haseebrabbani:
Two issues to fix:
1. **`ci.yml` coverage job: `--features guardian-server/postgres` removes the filesystem-backend tests from all of CI.** The coverage job was the only job running default-feature tests (clippy/build use `--all-features`, which also compiles them out). The five `#[cfg(not(feature = "postgres"))]` tests in `crates/server/src/builder/storage.rs` (~464–522) — including `filesystem_rejected_in_prod_stage` and `filesystem_build_emits_audit_startup_warning` (FR-021/SC-011) — now run in no job. Suggest a second no-postgres test invocation (or running llvm-cov twice) so the default backend stays covered.
2. **Newly-activated env-var tests race.** The three `test_resolve_pool_size_*` tests (`builder/storage.rs:615–656`) first run in CI because of this PR and mutate `GUARDIAN_DB_POOL_MAX_SIZE` via `set_var`/`remove_var` with no lock under the parallel runner — a latent flake in the coverage job. They should hold a lock per the `testing/env_lock.rs` convention (a module-local one suffices; the var is private to this module).

Minor stuff:
1. `testing/pg.rs` `create_database`: `set_path("/postgres")` keeps the query string, so a `dbname=` param (which the module documents as overriding the path) makes the maintenance connection target the same missing DB — auto-create can never work for that URL shape. Strip `dbname` from the maintenance URL.
2. `testing/pg.rs` probe: classification is English substring matching with a `Retryable` fallthrough, so e.g. `role "guardian" does not exist` or reworded pooler errors burn the full 30 s and die under a "did not accept connections" headline. Consider probing the maintenance DB + `SELECT 1 FROM pg_database` instead of parsing error text (SQLSTATE isn't available for connect-phase failures through diesel).
3. ~16 call sites still run `run_migrations(&url)` right after `test_database_url()`, which already migrated in `prepare()` — only the deliberate re-apply after the revert test needs to stay. Dropping the rest keeps the setup contract in one place.

haseebrabbani (approved):
lgtm, few nits (none blocking):

- README's "Rust Tests" section still lists only integration/e2e — worth adding the Postgres suite.
- The `T040 (FR-026 / SC-009)` traceability tags didn't survive the audit-trigger doc rewrite.
- A panic in the backfill test between `revert_last_migration` and the re-apply would leave the schema half-migrated for the rest of the run (the `OnceCell` won't re-reset); a drop-guard re-apply would contain it.
- Test selection now rests on the documented "module path contains `postgres`" convention — holds today (checked), but nothing compile-time enforces it for future tests.

Issue #357, opened 2026-07-28
Guardian based recovery does preserve the transaction history and also does not recover pending notes


0xnullifier:
If I recover my account using `/lookup`, it does not recover the pending notes (both public + private) , this could potentially be solved at a client level by syncing from the tip again (or the search could be narrowed somehow) but AFAIK this would take too long potentially hours.

The transaction history concern is a little less because we can get away with this for now (before-mainnet)

This is mainly solved with #307 where we can put the encrypted note bytes and transaction history but 307 is a major commitment infrastructure wise and I wonder if we would be better off with a wallet backend for some storage

haseebrabbani:
@0xnullifier we've done a design pass on this issue plus a feasibility spike (#412), and want to lay out the plan before implementation so you can provide any feedback.

## How we're breaking this down

The issue bundles several problems with very different costs, so we're splitting it:

1. **Transaction history** → separate issue. Guardian already retains every transaction summary per account indefinitely (`deltas` table) and can decode full note details from it — it's just only exposed on the operator dashboard today. A client-facing, authenticated, paginated history endpoint is a no-schema-change feature. Splitting it out so it ships on its own track rather than waiting on note recovery.

2. **Pending public notes** → recover from chain, client-side only. The spike (#412) showed tag-scoped `SyncNotes` from **genesis** is fast on testnet (1.6M blocks): ~0.3 s for a quiet tag, ~2 s for a tag with ~3k matches, 45 s worst-case for the densest tag on the network. Cost scales with matches, not chain length — so no server change and no stored block heights are needed for this. (The "takes hours" cost you observed comes from full `syncState()`, which does per-note store work the tag-scoped scan skips; fresh-store `syncState` with a busy tag exceeded 10 minutes in our measurement.)

3. **Pending private notes** → three recovery sources, best-effort, in priority order:
   - **Note-transport backlog drain** — validated live: a completely fresh store that tracks the account tag recovers a transport-delivered private note in ~1.3 s (`fetch_all_private_notes` exists upstream and can't regress the cursor). Limited by transport retention/pruning.
   - **Notes embedded in pending `consume_notes` proposals** — v2 proposals already carry serialized note bytes; the spike validated rebuilding an importable note from those bytes + a node-fetched inclusion proof into a fresh store (works for private notes too, since the node serves proofs without bodies).
   - **Wallet-owned encrypted backup bundle** — wallet-side, for durable recovery beyond transport retention. Guardian never sees keys.

[...]

PR #420, opened 2026-08-20
feat: client-facing transaction history endpoint over existing delta data (#413)

### LINKED ISSUE #413: Client-facing transaction history endpoint over existing delta data
Sub-issue of #357 (the "transaction history" half of its title).

## Context

Guardian already retains a complete per-account transaction history: the `deltas` table stores the raw `TransactionSummary` for every canonical nonce indefinitely, and the delta-summary projection (`decode_full` in `crates/server/src/delta_summary/projection.rs`) already decodes full note details from it — note ids, tags, assets, sender/recipient, and P2ID/P2IDE classification (including `reclaim_height`/`timelock_height` from #366).

Today that view is exposed only behind operator-dashboard auth (`GET /dashboard/accounts/{id}/deltas/{nonce}`). Wallets have no way to render an account's history after recovery.

The existing client-facing `/delta/since` does not cover this: it merges deltas into a single blob for state fast-forward and drops per-transaction boundaries and note detail.

## Scope

- A client-facing, cosigner-authenticated, **paginated** per-account history endpoint (HTTP + gRPC) over the existing `deltas` data, reusing the dashboard projection for note decoding.
- SDK methods to consume it (TypeScript + Rust, semantic parity).
- OpenAPI / proto updates and docs.

**No schema change and no new stored data** — this is exposure of data Guardian already holds.

## Out of scope

- Note recovery (covered by the sibling sub-issues of #357).
- History of transactions the account did not push through Guardian (Guardian never sees those; documented limitation).

## Acceptance criteria

- [ ] Endpoint returns per-transaction entries (nonce-ordered, paginated) with decoded input/output note summaries.
- [ ] Auth is the standard cosigner check; pause state does not block reads.
- [ ] TS and Rust SDK methods with matching semantics.
- [ ] Docs updated (`docs/MULTISIG_SDK.md`, spec/api.md).

No dependencies — can proceed in parallel with the recovery primitives.

haseebrabbani:
Closes #413 (sub-issue of #357, the "transaction history" half).

## What

Adds a client-facing, cosigner-authenticated, paginated per-account delta history feed over the delta data Guardian already retains — no schema change, no new stored data:

- **HTTP** `GET /delta/history?account_id=&limit=&cursor=` and **gRPC** `GetDeltaHistory`, both thin transports over a new `services::get_delta_history`.
- **SDK methods** with matching semantics: Rust `GuardianClient::get_delta_history` + `MultisigClient::delta_history()` (typed `HistoryPage`), TS `GuardianHttpClient.getDeltaHistory` + `Multisig.deltaHistory()`.
- OpenAPI regenerated; `spec/api.md`, `docs/MULTISIG_SDK.md`, and all four package/crate READMEs document the new surface.

## Semantics

- **Canonical deltas only**, newest-first by nonce. Candidates are in flight and retained/discarded rows never became account state; pending proposals stay on the proposal endpoints. Named `/delta/history` (not a transaction-specific path) so a future `status=` filter can expose other lifecycle statuses, defaulting to canonical — see review discussion.
- Each entry carries `nonce`, an explicit `status` (always `canonical` today; carried so the future filter is a pure widening), canonical `timestamp`, `new_commitment` (for correlating with recovered state), and input/output note summaries decoded by the existing dashboard projection (`decode_transaction_summary` → `decode_full`) — note id, P2ID/P2IDE/swap/mint/burn classification, `note_type` (`public` / `private` from the on-chain note metadata), assets, sender/recipient where the note script exposes them. An undecodable payload returns the entry with empty sections plus a `decode_warnings` item (dashboard-detail parity), never a 500.
- **First paginated client endpoint**: reuses the dashboard pagination infra (`PagedResult`, HMAC cursors, limit 50 default / 500 max, `invalid_limit` / `invalid_cursor`). Cursors use a dedicated `AccountDeltaHistory` kind and are **bound to the minting account**, so they can be replayed neither across surfaces nor across accounts. The multi-replica cursor-secret warning and `docs/CONFIGURATION.md` / scaling guide now cover the client feed.
- Auth is the standard `resolve_account` cosigner check; **pause state does not block the read** (integration-tested). EVM accounts are rejected with `unsupported_for_network` like the other Miden delta APIs (documented in the endpoint's 400 response).
- Signed request payload: canonical JSON of the query struct with `limit` kept a string and omitted keys skipped; TS tests assert byte-match. gRPC signs the encoded message as usual (client proto mirrors field numbers exactly).

## Storage

New `StorageBackend::list_canonical_deltas_paged` (Postgres pushes `status_kind = 'canonical'` + nonce cursor into SQL; filesystem filters in memory like its sibling; both decorators and the mock covered). Cursor keys on the immutable `(account_id, nonce)` unique, so traversal is fully stable.

## SDK typing

- TS exposes closed unions for the wire vocabularies (`HistoryNoteTag`, `HistoryNoteVisibility`, `HistoryDecodeSection`, `HistoryEntryStatus`).
- The Rust multisig SDK exposes the same vocabularies as typed enums with an `Other(String)` fallback, so unknown future wire labels degrade gracefully instead of erroring.
- guardian-operator-client mirrors the shared `DecodedNote`'s new `note_type` as an optional field.

## Also in this PR

- CI: the miden-multisig-client npm test job now builds and overlays the in-repo `guardian-client` before testing, so cross-package contract PRs (the default per AGENTS.md) are tested against the pair that ships together instead of the last published registry version.
- A kind-valid cursor with no resume position is rejected instead of silently restarting at page 1.
- `parse_limit` now delegates its range policy to a new `validate_limit` (used by the gRPC transport, which carries `limit` as `optional uint32`).
- Wire-label parity tests pinning the `as_str()` gRPC labels to the serde (JSON) labels for `NoteTag` / `AssetKind` / `DecodeSection` / `NoteVisibility` / `HistoryEntryStatus`.
- `KNOWN_GRPC_METHODS` metrics allowlist gained `GetDeltaHistory` (test-enforced).
- Examples: the Rust demo gained a "Delta history" menu action (interactive page walk) and the browser example a self-contained `DeltaHistoryCard` on the multisig dashboard (load/refresh + cursor-based "Load more"); the web example's `initMultisigClient` was synced with the `_shared` copy (missing `rpc` param) so the example typechecks again.

## Known limitations / follow-ups (not in scope)

- Output notes carried as partial notes in the stored summary (e.g. private notes) classify as `custom` with no recipient — pre-existing projection behavior, documented in the SDK docs (their `note_type` is still reported).
- Only transactions pushed through Guardian appear (documented limitation from the issue).
- Candidates for later: the `status=` filter (defaulting to canonical) from the review discussion — at which point `list_canonical_deltas_paged` collapses into a status-filtered variant of `list_account_deltas_paged`; a shared pagination scaffold across the six paginated services; a Postgres partial index on `(account_id, nonce DESC) WHERE status_kind = 'canonical'` if proposal-churn-heavy accounts show up


zeljkoX (commented):
LGTM

I will continue reviewing.

Posting initial review.

Mainly around naming and future extensions. 

I believe we should leave door open for extending with other statuses. For now we can support just canonical, but in the future we should expose all like discarded, candidate retained and discarded.

In order to support that in the future I believe fe can use something as:

- HTTP | GET /delta/history
- gRPC | GetDeltaHistory

Default query stays canonical-only (issue #413). Later:

- GET /delta/history?account_id=…&limit=50&status=canonical
- GET /delta/history?account_id=…&status=canonical,discarded

This would also mean propagating different naming to files and variables. For example:
- get_transaction_history.rs -> get_delta_history.rs
- history_grpc.rs -> delta_history_grpc.rs
- history.rs -> delta_history.rs
- ....

Interested in your opinion, let me know if this makes sense to you.

haseebrabbani:
@zeljkoX yea I went back and forth on naming, but I like what you proposed with `GET /delta/history` and potential status filtering. will rename endpoints/methods

crates/server/src/dashboard/cursor.rs
```diff
@@ -84,6 +84,12 @@ pub enum CursorKind {
     AccountProposals,
     GlobalDeltas,
     GlobalProposals,
+    /// Client-facing per-account canonical history feed (issue #413).
+    /// Carries `last_nonce` plus `last_account_id` (the account the
+    /// cursor was minted for), and is its own kind so dashboard and
+    /// client cursors cannot be replayed across surfaces — nor a
+    /// history cursor across accounts.
+    AccountHistory,
```
zeljkoX: nit: AccountDeltaHistory

packages/miden-multisig-client/src/multisig.ts
```diff
@@ -1166,6 +1166,19 @@ export class Multisig {
...
+  /**
+   * Fetch one page of this account's canonical transaction history
+   * from GUARDIAN (issue #413), newest-first by nonce, with decoded
+   * input/output note summaries. Pass `options.cursor` from a previous
+   * page's `nextCursor` to resume; an absent `nextCursor` means the
+   * feed is exhausted. Served while the account is paused. Only
+   * transactions pushed through GUARDIAN appear — history of
+   * transactions executed elsewhere is not visible to it.
+   */
+  async deltaHistory(options: HistoryOptions = {}): Promise<HistoryPage> {
```
zeljkoX, line 1230: Let's update all client Readme files to document new method

docs/MULTISIG_SDK.md
```diff
@@ -537,6 +537,39 @@ support it. Multiple matches are valid: the same key commitment may authorize
 more than one account, and the method returns all matches instead of choosing
 one implicitly.
 
+### Transaction History
```
zeljkoX: Maybe to use Delta History term here and in other docs

 ### Delta History
Guardian retains the account's canonical delta history, allowing a wallet to
render its transaction history after recovery. `deltaHistory()` returns one
page at a time, newest-first by nonce, with server-decoded input and output note
summaries: note ID, P2ID/P2IDE/swap/mint/burn classification, assets, and
sender or recipient when exposed by the note script.

crates/server/src/services/get_delta_history.rs
```diff
@@ -0,0 +1,282 @@
...
+    pub limit: u32,
+    pub cursor: Option<Cursor>,
+    pub credentials: Credentials,
+}
+
+/// One canonical transaction in the history feed wire shape.
+/// `account_id` is omitted — the authenticated query scopes it.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, utoipa::ToSchema)]
+pub struct HistoryEntry {
+    pub nonce: u64,
```
zeljkoX, line 38: maybe to include status even if now it is canonical for every entry. In the future once we add filtering support it would mean less model changes.

crates/server/src/services/get_delta_history.rs
```diff
@@ -0,0 +1,282 @@
...
+    pub account_id: String,
+    pub limit: u32,
+    pub cursor: Option<Cursor>,
+    pub credentials: Credentials,
+}
+
+/// One canonical transaction in the history feed wire shape.
+/// `account_id` is omitted — the authenticated query scopes it.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, utoipa::ToSchema)]
+pub struct HistoryEntry {
```
zeljkoX: Maybe to add `proposal_type` like we have in dashboard deltas endpoint. It could give more context to users.

zeljkoX, line 37: What about adding field `note_type: private | public` for DecodedNote. This can be obtained from metadata.note_type().

For private notes for privacy reasons maybe to exclude recipient field.

zeljkoX (approved):

Looks good.

Added few more comments. Also there are few Copilot ones to tackle before merging.

crates/server/src/services/get_delta_history.rs
haseebrabbani: Added `note_type: "public" | "private"` from `metadata().note_type()` but kept `recipient` for private notes: the caller is an authenticated cosigner who can already fetch the full note from `GET /delta`'s payload, so omitting it here wouldn't add privacy, just degrade wallet UX. Happy to drop it if you see a leak vector I'm missing

Issue #412, opened 2026-08-18
Spike: note-recovery feasibility benchmarks (genesis tag-scan, transport drain, proposal-note reconstruction)


haseebrabbani:
Time-boxed feasibility spike (~2 engineering days) for the note-recovery work planned under #357. Everything downstream of the recovery plan is shaped by these answers, so this runs first and the other sub-issues don't start until it lands.

## Context

The #357 implementation plan recovers pending notes from four sources after a Guardian lookup: private-note transport backlog, notes embedded in pending `consume_notes` proposals, an encrypted wallet bundle, and historical public-note discovery by account tag. Its v1 lower bound for historical discovery is **genesis**, with an account creation-block anchor as a possible later optimization. Before committing to that shape, we need to validate the load-bearing assumptions and measure whether genesis-to-tip tag scans are acceptable.

### 1. Historical public-note discovery cost (the go/no-go)

Benchmark tag-scoped note sync (`SyncNotes`) from genesis to tip against a real network (testnet), for a single account tag:

- Full genesis → tip scan: wall-clock time, page count, RPC round-trips.
- Incremental ranges: 1k, 10k, 100k blocks.
- Zero-match scan vs. sparse matches vs. many matches.
- Compare against a normal `syncState()` from a clean store (the status-quo "hours" path) for reference.

### 2. WASM `syncNotes` pagination coverage

Confirm the TS/WASM `syncNotes` surface covers an arbitrary block range: does it paginate internally to the requested `toBlock`, or return bounded pages requiring client-side chunking? Verify with an integration test over a range wider than one page.

### 3. Proposal-embedded note reconstruction

Validate that a serialized `Note` from a v2 `consume_notes` proposal (`consume_notes_notes`) can be rebuilt into an importable note:

- Decode base64 → `Note`, fetch the current inclusion proof from the node, construct `NoteFile::NoteWithProof` (and the TS equivalent), import individually.
- Confirm this works for **private** notes (node has no body — must combine local bytes + node proof) and public notes.
- Confirm behavior for a not-yet-committed note (expect a clean, retryable failure, not an abort).

### 4. Transport full-drain semantics in a dirty shared store

Using upstream `fetch_all_private_notes()` (Rust) / `fetchPrivate({ mode: 'all' })` (TS):

- Confirm a full drain re-scans from the beginning without regressing the persisted transport cursor.
- Confirm idempotence (second drain imports nothing new).
[...]

PR #423, opened 2026-08-20
feat(sdk): private-note transport backlog drain recovery primitive (#414)

### LINKED ISSUE #414: SDK recovery primitive: private-note transport backlog drain
Sub-issue of #357. Spike #412 validated the underlying mechanics; this issue productizes them as an SDK recovery primitive.

## Context

After recovery on a fresh device, the local store has no note-transport cursor — and in a shared dirty store, another account's sync may have advanced the cursor past notes belonging to the newly recovered account. Recovery therefore needs a full transport rescan.

Spike findings (#412):
- Full device-loss round trip verified live against the testnet transport: a fresh store that tracks the account tag recovers a transport-delivered private note via one `fetch_all_private_notes()` drain in ~1.3 s.
- Drains are idempotent; the incremental fetch afterwards is a no-op.
- The upstream drain cannot regress the stored cursor by construction (it persists `max(drain_cursor, stored_cursor)`).
- The drain is **tag-scoped**: a store with no tracked tags drains nothing. The account's note tag must be tracked before draining.
- Transport delivery requires no on-chain transaction (notes arrive `Expected` until observed on chain).

## Scope

- TS: `drainPrivateNoteBacklog(midenClient)` wrapping `fetchPrivate({ mode: 'all' })`.
- Rust: `drain_private_note_backlog()` on `MultisigClient`, wrapping `fetch_all_private_notes()`.
- Structured `TransportRecoveryReport { status: completed | unavailable | failed, imported, reason? }`. Transport-disabled / unreachable must be reported, not thrown — a transport failure must not abort the rest of a recovery flow.
- Catch the upstream convergence-guard error (`PaginationDidNotTerminate`, 1000 iterations) and surface it as a retryable failure.
- **Load/tag regression tests** (these gate whether the drain sees the account at all): `load()` (TS) and `pull_account` (Rust) insert the recovered account, its standard note tag is behaviorally tracked afterwards, and reloading an existing account stays idempotent.
- Rename the misleading Rust helper argument `imported` → `overwrite` while in the area.

## Acceptance criteria

- [ ] Both SDKs with semantic parity; report types as above.
- [ ] Tests: idempotence, tag-scoping (no tag → nothing drained), cursor non-regression, unavailable-transport reporting, load/tag regression suite.
- [ ] Docs: transport recovery is bounded by transport retention/pruning — it is **not** a backup (senders may not use transport; blobs get pruned).

Unblocked (spike complete). Ordering: pairs naturally with the other two recovery primitives from the #357 plan.

haseebrabbani:
Closes #414 (sub-issue of #357). Productizes the transport-drain mechanics validated by spike #412 as an SDK recovery primitive in both SDKs.

## What

- **TS**: `drainPrivateNoteBacklog(midenClient)` wrapping `fetchPrivate({ mode: 'all' })`, exported from the package root. Pass the same `MidenClient` instance that was injected into `MultisigClient` — the drain only sees note tags tracked in that store.
- **Rust**: `MultisigClient::drain_private_note_backlog()` wrapping `fetch_all_private_notes()`.
- Structured `TransportRecoveryReport { status: completed | unavailable | failed, imported, retryable, reason? }`. Transport problems are reported, never thrown, so a transport failure cannot abort the rest of a recovery flow; an error from the call itself means the local store failed.
- Rename of the misleading `add_or_update_account` argument `imported` → `overwrite` (matches the upstream `add_account` parameter it forwards to; the doc now also notes the flag is currently inert because the existence pre-check routes tracked accounts through the always-overwrite path).
- Docs in `docs/MULTISIG_SDK.md` and both package READMEs, including the retention caveat: transport recovery is **not a backup** (senders may bypass the transport; blobs are pruned).

## Classification semantics

- `unavailable` guarantees `imported == 0`: transport disabled, or unreachable before anything landed. A connection lost mid-drain **after** partial progress reports `failed` + `retryable` with the partial count kept — rerunning continues it (imports are idempotent).
- The `retryable` bit reuses the crate's existing transient/permanent classifiers: `is_transient_note_transport_error` inspects the `Connection` cause chain (endpoint-parse/TLS misconfig is not retryable), `is_transient_rpc_error` decides for node RPC failures mid-import. The upstream `PaginationDidNotTerminate` convergence guard is surfaced as a retryable failure.
- The Rust `NoteTransportError` match is deliberately exhaustive, so a new upstream variant is a compile error rather than a silently wrong default.
- TS classification is message-fragment based (no typed errors cross the WASM boundary). The fragments are pinned against the shipped WASM binary's string table in `tests/recovery-fragments.test.ts`, so an SDK bump that rewords the upstream error text fails CI instead of silently misclassifying.

## Notes

- `imported` is a store-length delta in both SDKs: upstream `fetch_all_private_notes` computes but discards the imported note IDs, so concurrent store activity during a drain is attributed to it (documented). Having upstream return the count would let both wrappers drop the before/after scans — candidate for an upstream issue.

### DIFF
--- a/crates/miden-multisig-client/README.md
+++ b/crates/miden-multisig-client/README.md
@@ -179,6 +179,35 @@ client.sign_proposal(&to_sign.id).await?;
 client.execute_proposal(&proposal.id).await?;
 ```
 
+### Draining the Private-Note Transport Backlog
+
+After recovery on a fresh device the local store has no note-transport
+cursor — and in a store shared with other accounts, sync may have advanced
+the cursor past private notes addressed to the newly recovered account.
+`drain_private_note_backlog` rescans the full transport backlog for every
+tracked note tag, regardless of the stored cursor. Run it after
+`pull_account`: the drain only sees tags tracked in the store, and
+`pull_account` inserting the account is what tracks its tag.
+
+```rust
+client.pull_account(account_id).await?;
+
+let report = client.drain_private_note_backlog().await?;
+// report.status: Completed | Unavailable | Failed
+// report.imported: number of newly imported note records
+```
+
+Transport problems are reported in the result, never returned as errors:
+`Unavailable` means no transport endpoint is configured or the transport
+could not be reached; `Failed` with `retryable: true` keeps any partial
+progress and rerunning the drain continues it. The drain is idempotent and
+never regresses the stored transport cursor.
+
+Transport recovery is **not a backup**: it is bounded by the transport
+service's retention. Senders may deliver private notes out-of-band without
+using the transport, and relayed blobs are pruned after the retention
+window.
+
 ### Recovering From a Dead Transaction (Abandon)
 
 If `execute_proposal` dies after guardian approval (RPC submit failure,
--- a/crates/miden-multisig-client/README.md
+++ b/crates/miden-multisig-client/README.md
@@ -199,9 +199,10 @@ let report = client.drain_private_note_backlog().await?;
 
 Transport problems are reported in the result, never returned as errors:
 `Unavailable` means no transport endpoint is configured or the transport
-could not be reached; `Failed` with `retryable: true` keeps any partial
-progress and rerunning the drain continues it. The drain is idempotent and
-never regresses the stored transport cursor.
+could not be reached before anything was imported; `Failed` with
+`retryable: true` keeps any partial progress and rerunning the drain
+continues it. The drain is idempotent and never regresses the stored
+transport cursor.
 
 Transport recovery is **not a backup**: it is bounded by the transport
 service's retention. Senders may deliver private notes out-of-band without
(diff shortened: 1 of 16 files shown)


crates/miden-multisig-client/README.md
```diff
@@ -179,6 +179,36 @@ client.sign_proposal(&to_sign.id).await?;
 client.execute_proposal(&proposal.id).await?;
 ```
 
+### Draining the Private-Note Transport Backlog
```
zeljkoX, line 182: not sure how this documentation would loo like once all options lands. would there be single recovery method where user could select source/s?

zeljkoX (approved):
Let's merge this against feature branch so we can have one final review once all recovery options land so we can see how orchestration between works.

crates/miden-multisig-client/README.md
haseebrabbani: The primitives stay separate building blocks (they have different inputs: the drain needs only the loaded account, proposal import consumes the synced proposals, backfill takes an RPC endpoint/range, and deliberately different report shapes), but once all PRs land a thin `recoverNotes({ sources? })` orchestrator that runs all of them and returns a combined report is a natural addition for the common "just recover everything" case. In that case, these README sections collapse into one "Recovering notes" section. We can decide that in the final review on the feature branch.

Issue #416, opened 2026-08-19
SDK recovery primitive: historical public-note backfill by tag


haseebrabbani:
Sub-issue of #357. Spike #412 benchmarked the approach and confirmed it is fast without any server-side changes.

## Context

Public notes addressed to an account are on-chain, but normal sync starts from the store's **global** cursor. In a shared dirty store the cursor may already be past blocks containing a recovered account's notes (`H < B ≤ T`), and a fresh store has no efficient path at all — full `syncState()` is the "hours" path #357 describes.

Spike findings (#412), testnet at ~1.61M blocks:
- Genesis→tip tag-scoped `SyncNotes` for one tag: **0.34 s** zero-match, **2.0 s** at ~3k matching blocks, **44.5 s** at 135k matching blocks. Cost scales with matches, not range length.
- For contrast: fresh-store `syncState()` with that dense tag took **11.8 hours** (~950× slower).
- Both the Rust and WASM/TS `syncNotes` paginate the full requested range internally — no client-side chunking needed.
- Safety cap: miden-client errors (`PaginationError`) after 1000 pages per tag-chunk rather than truncating.
- The creation-block anchor is therefore **optional** at current network scale (tracked separately; see the anchor issue) — genesis is an acceptable v1 lower bound and this primitive is unblocked now.

## Scope

- TS: `backfillPublicNotesByTag(midenClient, { accountId, midenRpcEndpoint, fromBlock = 0, toBlock = tip })` using the raw WASM `syncNotes`.
- Rust: `backfill_public_notes_by_tag(account_id, from_block, to_block)` on `MultisigClient` using the existing `sync_notes` wrapper.
- Build the account's standard tag (`NoteTag::with_account_target`), scan the range, keep public notes, fetch bodies/proofs, import **individually** as `NoteWithProof`.
- **Never mutate the global sync height** — normal forward sync runs afterwards.
- Catch `PaginationError` and fall back to client-side range splitting; report, don't fail recovery.
- `PublicBackfillReport { scannedFrom, scannedTo, discovered, outcomes }`.

## Acceptance criteria

- [ ] Both SDKs with semantic parity.
- [ ] The dirty-store scenario test: coverage at `H`, note committed at `B`, global cursor at `T` with `H < B ≤ T` — recovery discovers the note at `B` without rewinding `T`.
- [ ] Duplicate results and tag collisions are tolerated (tags are best-effort filters).
- [ ] Pagination-cap fallback covered by a test.
[...]

PR #424, opened 2026-08-20
feat(sdk): import proposal-embedded notes as a recovery primitive (#415)

### LINKED ISSUE #415: SDK recovery primitive: proposal-embedded note import
Sub-issue of #357. Spike #412 validated the reconstruction path; this issue productizes it as an SDK recovery primitive.

## Context

v2 `consume_notes` proposals embed serialized note bytes (`consume_notes_notes`, base64 `Note` serialization, introduced by #229). After recovery, pending proposals are already restored via `syncProposals()` — the embedded note bytes are opportunistic recovery material for notes the account was in the middle of consuming.

Spike findings (#412), on a fresh store against testnet:
- Local note bytes + node-fetched inclusion proof → `NoteFile::NoteWithProof` import succeeds without using any node-held body; the record lands in `Unverified` state and the next sync verifies it. This works for **private** notes too — the node serves inclusion proofs without bodies (`FetchedNote::Private`).
- An uncommitted note returns 0 results from `get_notes_by_id` — cleanly classifiable as `not-committed`/retryable, no error — and can be parked in `Expected` state via a `NoteDetails` import.
- `import_notes` batches are **atomic** upstream: one invalid note fails the whole batch. Imports must be per-note.

## Scope

- TS: `importNotesFromProposals(midenClient, proposals, { midenRpcEndpoint })`.
- Rust: `import_notes_from_proposals(&[Proposal])` on `MultisigClient`.
- Per note: decode base64 → deserialize `Note` → check already-present/consumed → fetch inclusion proof → import individually as `NoteWithProof` (or `NoteDetails`/`Expected` for not-yet-committed) → outcome.
- Per-note `NoteImportOutcome { identifier, source: 'proposal', status: imported | already-present | already-consumed | not-committed | invalid | failed, retryable?, reason? }`. A malformed note must not block the others.

## Acceptance criteria

- [ ] Tests: valid public, valid private (body from local bytes only), uncommitted (retryable, no abort), malformed (isolated failure), already-consumed classification.
- [ ] Docs: proposals are opportunistic recovery material, not a backup (may be absent, stale, or deleted after canonicalization); embedded note bytes are visible to the Guardian operator — existing v2 behavior, not a new exposure.

Unblocked (spike complete).

haseebrabbani:
Closes #415 (sub-issue of #357). Productizes the reconstruction path validated by spike #412.

## What

Adds a recovery primitive to both SDKs that rebuilds Miden store records from the note bytes embedded in v2 `consume_notes` proposals (#229), plus a node-fetched inclusion proof — so it works for private notes without the node ever holding the body.

- Rust: `MultisigClient::import_notes_from_proposals(&[Proposal]) -> Vec<NoteImportOutcome>` (new `client/recovery.rs`)
- TS: standalone `importNotesFromProposals(midenClient, proposals, { midenRpcEndpoint, rpc? })` (new `src/recovery.ts`)

Per unique embedded note (duplicates across proposals fold into one outcome): decode → skip notes the store already tracks → one batched inclusion-proof fetch → import individually (upstream `import_notes` batches are atomic, so one bad note must not sink the rest) → per-note `NoteImportOutcome` with `status` ∈ `imported | already-present | already-consumed | not-committed | invalid | failed` and a `retryable` flag. No per-note problem aborts the batch; the Rust method returns a plain `Vec` because nothing can fail the whole call.

## Key behaviors (live-tested against testnet)

- **Not-committed notes** are recorded in `Expected` state *with their tag tracked* so a later sync picks them up once they commit; reported `not-committed`/retryable. On the TS side the WASM details file cannot carry a tag, so the tag is registered via `addTag` before the import — tag-first so a failure never leaves an untagged dead record.
- **Chain-nullified notes**: miden-client stores these as metadata-less consumption-history records; the primitive reports them `already-consumed` instead of pretending they were recovered.
- **Metadata-less store records** (details-only expected imports, consumed-external history) expose neither a note ID nor a nullifier, so existing-record matching uses details commitments (Rust) / recipient digests (TS — the WASM `NoteFilter` has no details-commitment variant); an ID-only lookup would re-import them forever.
- **Retryability** is classified, not assumed: transient RPC failures during the proof fetch or the import's internal node calls are `retryable: true`.
- Decoding is deliberately permissive: the strict `note.id() == note_ids[i]` binding check stays in the verify/execute path; embedded notes are self-validating, so recovery imports whatever real notes the bytes decode to.

## Docs

`docs/MULTISIG_SDK.md` (both language sections + API tables) and both package READMEs, per AGENTS.md §9. Example wiring (demo/smoke-web) is deferred to the #357 recovery-flow orchestration, where the three primitives (#414/#415/#416) compose.

## Notes for reviewers

- The TS entry point is a standalone function because the issue specifies that signature; a `Multisig` convenience method delegating to it would be a natural follow-up.
- The TS pre-import store scan reads all input notes once (WASM filter surface limitation, commented in code); extracting a shared matching helper is planned with #416.

### DIFF
--- a/docs/MULTISIG_SDK.md
+++ b/docs/MULTISIG_SDK.md
@@ -631,27 +631,28 @@ export/import helpers).
 
 After key-based recovery the local Miden store starts empty, so notes the
 account was in the middle of consuming are gone. v2 `consume_notes` proposals
-embed the serialized notes they consume (issue #229), and the standalone
-`importNotesFromProposals` helper turns those embedded bytes back into store
-records (issue #415): it fetches each note's on-chain inclusion proof and
-imports the note individually, so it works for private notes too — the node
-never needs to hold the note body.
+embed the serialized notes they consume (issue #229), and
+`multisig.importNotesFromProposals()` turns those embedded bytes back into
+store records (issue #415): it fetches each note's on-chain inclusion proof
+and imports the note individually, so it works for private notes too — the
+node never needs to hold the note body. The method reuses the client's Miden
+RPC endpoint and retry configuration, and syncs pending proposals from
+GUARDIAN when none are passed.
 
 ```typescript
-import { importNotesFromProposals } from '@openzeppelin/miden-multisig-client';
-
 const multisig = await client.load(accountId, signer);
-const proposals = await multisig.syncProposals();
 
-const outcomes = await importNotesFromProposals(midenClient, proposals, {
-  midenRpcEndpoint: 'https://rpc.testnet.miden.io',
-});
+const outcomes = await multisig.importNotesFromProposals();
 for (const outcome of outcomes) {
   console.log(outcome.identifier, outcome.status, outcome.reason ?? '');
 }
 
 await multisig.syncState(); // verifies the imported notes
 ```
+
+A standalone `importNotesFromProposals(midenClient, proposals, {
+midenRpcEndpoint, rpc? })` export backs the method for callers holding a raw
+WASM client or proposals from another source.
 
 Each unique embedded note gets its own `NoteImportOutcome` with a `status` of
 `imported`, `already-present`, `already-consumed`, `not-committed`, `invalid`,
@@ -902,6 +903,7 @@ Standalone functions:
 | `exportNoteToFile(noteId, filename?)` | Browser-only: download the note file |
 | `importNoteFromBytes(noteBytes)` | Import a note file received out-of-band |
 | `importNoteFromFile(file)` | Import a note file from a browser `File`/`Blob` |
+| `importNotesFromProposals(proposals?)` | Import notes embedded in pending v2 consume-notes proposals, reusing the client's RPC settings; syncs proposals when none are passed (issue #415) |
 | `createAddSignerProposal(commitment, { nonce, newThreshold }?)` | Create add signer proposal (`newThreshold` defaults to the current threshold) |
 | `createRemoveSignerProposal(commitment, { nonce, newThreshold }?)` | Create remove signer proposal (`newThreshold` defaults to min of current threshold and remaining signer count) |
 | `createChangeThresholdProposal(threshold, { nonce }?)` | Create threshold change proposal |
--- a/docs/MULTISIG_SDK.md
+++ b/docs/MULTISIG_SDK.md
@@ -627,6 +627,46 @@ transport, and relayed blobs are pruned after the retention window. Notes
 outside both are recoverable only from their sender (see the note
 export/import helpers).
 
+### Recovering Notes From Pending Proposals
+
+After key-based recovery the local Miden store starts empty, so notes the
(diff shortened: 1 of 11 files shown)


zeljkoX:
Does it makes sense to create PRs against feature branch? Given few recovery strategies it would be good to have one final review on all included and orchestration between them.

packages/miden-multisig-client/README.md
```diff
@@ -171,6 +171,36 @@ several accounts, and the helper surfaces all of them rather than silently
...
+standalone `importNotesFromProposals` helper turns those embedded bytes back
+into store records: it fetches each note's on-chain inclusion proof and
+imports the note individually, so it works for private notes too.
+
+```typescript
+import { importNotesFromProposals } from '@openzeppelin/miden-multisig-client';
+
+const proposals = await multisig.syncProposals();
+const outcomes = await importNotesFromProposals(midenClient, proposals, {
+  midenRpcEndpoint: 'https://rpc.testnet.miden.io',
```
zeljkoX: Why rpc url is needed here? multisig should already have info about it.

docs/MULTISIG_SDK.md
```diff
@@ -537,6 +537,46 @@ support it. Multiple matches are valid: the same key commitment may authorize
...
+imports the note individually, so it works for private notes too — the node
+never needs to hold the note body.
+
+```typescript
+import { importNotesFromProposals } from '@openzeppelin/miden-multisig-client';
+
+const multisig = await client.load(accountId, signer);
+const proposals = await multisig.syncProposals();
+
+const outcomes = await importNotesFromProposals(midenClient, proposals, {
```
zeljkoX: probably there should be single exposed recovery method in multisig where user can pass recovery strategies he wants to use

packages/miden-multisig-client/README.md
zeljkoX: we should avoid needing to pass rpc url, it should be reused from client,

zeljkoX: rust uses client method, ts uses non client method

let proposals = client.list_proposals().await?;
let outcomes = client.import_notes_from_proposals(&proposals).await;

zeljkoX: Double checking, is this going to be public api? Users probably should not set rpc on method level but on client level and reuse here. Setting rpc url at client level is already supported.

docs/MULTISIG_SDK.md
haseebrabbani: agreed, will handle this in final review of base feature branch

packages/miden-multisig-client/README.md
haseebrabbani: fixed

zeljkoX (approved):
LGTM

Thanks

PR #427, opened 2026-08-24
perf: build dependencies optimized for tests; run the suite via nextest

haseebrabbani:
## Problem

CI wall clock was the Test Suite job: 20m42s, of which ~17 minutes was pure test execution, not compilation. The server lib suite alone ran 896 tests in 828s — roughly 3.7 core-seconds per test, which pointed at the crypto dependencies (Falcon, RPO, Merkle) executing at `opt-level 0` under the dev/coverage profile.

## Changes

- **`[profile.dev.package."*"] opt-level = 3, debug = false`** — dependencies build optimized (and without debuginfo, which was the bulk of target-dir size) in dev/test/coverage builds; workspace crates keep `opt-level 0` for fast rebuilds, debugging, and coverage instrumentation. Dependencies compile once per lockfile/manifest change and stay cached. Every dev-profile CI job inherits the speedup.
- **`cargo nextest` for the test suite** — schedules every binary's tests in one parallel pool (plain `cargo test` parallelizes only within a binary) and prints per-test timings, so slow tests stay visible in CI logs.
- **Coverage on main pushes only** — instrumentation slows test execution 1.5–2× and recompiles workspace crates under coverage flags, and was the dominant remaining cost on the PR path. PRs run the identical test set uninstrumented via `cargo nextest run`; `cargo llvm-cov nextest` + the Codecov upload run on pushes to main, so Codecov keeps tracking trends. Doctests still run everywhere. Gating uses step-level `if:`, so skipped steps never leave required checks hanging.
- **Release Build job stays on every PR** (review feedback) — it is the only job compiling `--all-features` and the only release-profile (LTO) compile, so it is the sole pre-merge signal for feature-unification and link-stage breakage.
- **Caching reworked around the 10GB Actions cache quota** — per-job `Swatinem/rust-cache` (prunes workspace artifacts, dedupes target dirs), `CARGO_INCREMENTAL=0` workflow-wide, and docker-publish's buildkit cache moved from `mode=max` to `mode=min` (release-only workflow; frees ~2.4GB of intermediate layers).
- **Integration/Postgres job timeouts raised 20 → 35 minutes** — a cold dependency rebuild (any future `Cargo.lock` bump) must fit inside the timeout, and a timed-out job never runs its cache-save post step, which would otherwise loop cold forever.

## Measured impact

Verified on this PR's own runs with warm caches, all jobs green: **PR wall clock dropped 20m42s → ~5m** (the release Build job is the pole; test and lint feedback lands at ~3m22s). Per job: Test Suite 20m42s → 3m22s, Server Integration and E2E 7m52s → 2m03s, OpenAPI Spec Drift 5m30s → 3m03s, Authz Probe 4m58s → 1m29s, TypeScript multisig tests 4m41s → 1m39s, Contract Behavior 2m34s → 1m34s; Postgres and Clippy hold at ~2m; Build ~5m. Main pushes additionally run coverage (~8m, non-blocking). The cache store sits at ~7GB of the 10GB quota with zero churn across warm runs.

Correctness validation: the full workspace suite (1208 tests, same exclusions as CI) passes under nextest's process-per-test isolation, with the 2 known ignored Postgres tests skipped. The slowest remaining tests are contract-behavior VM executions at 2–4.6s each — genuine work, so no test-level changes were made. The Postgres suite job stays on `cargo test` (shared-database timing assumptions).

## Cache design notes (for future maintainers)

- Per-job rust-cache keys are deliberate: cache *sharing* between jobs loses to cargo feature unification — only the first job to save a shared key wins it, and every other job permanently rebuilds a multi-minute dependency delta (measured 6–12m per run in both a single-shared-key and a pairwise-shared experiment during this PR). Per-job keys were the only configuration observed uniformly warm.
- The first run after any `Cargo.lock`/workspace-manifest change re-primes dependencies for the affected jobs (~15m each, in parallel) before re-saving; subsequent runs are warm. After merge, main's first push pays one prime and saves main-scoped caches that all future PRs then restore from.
- Keep an eye on the cache store if new workflows start caching heavily — LRU eviction at the 10GB quota silently reverts jobs to cold rebuilds, which presents as "CI got slow again".

## Deferred follow-ups

- Skip the Rust jobs on docs-only PRs (needs a change-detector job so required checks report as skipped rather than hang).
- A merge queue would let the Build job run once per merge candidate instead of per commit, and validate PRs against latest main; requires admin enablement and a team process change, so out of scope here.
- Larger runners would roughly halve the remaining test execution if ~5m ever feels slow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,15 +57,31 @@ jobs:
           cache-on-failure: true
 
       - name: Install cargo-llvm-cov
+        if: github.event_name == 'push'
         uses: taiki-e/install-action@4a44d353abec8bd780e47de38e5bc5ea77053391 # cargo-llvm-cov
 
       - name: Install cargo-nextest
         uses: taiki-e/install-action@742a3ac9a75553d78e25773da9e6c438fdc5a74a # nextest
 
-      # nextest schedules every binary's tests in one parallel pool (cargo
-      # test parallelizes only within a binary) and prints per-test timings,
-      # so slow tests are visible in the log.
-      - name: Run tests with coverage
+      # Coverage runs on pushes to main only: instrumentation slows test
+      # execution 1.5-2x and forces workspace crates to recompile under
+      # coverage flags, and was the dominant remaining cost on the PR path.
+      # PRs run the same test set uninstrumented; Codecov tracks trends from
+      # main. nextest schedules every binary's tests in one parallel pool
+      # (cargo test parallelizes only within a binary) and prints per-test
+      # timings, so slow tests are visible in the log.
+      - name: Run tests (PRs, uninstrumented)
+        if: github.event_name != 'push'
+        run: >-
+          cargo nextest run
+          --workspace
+          --exclude guardian-prod-benchmarks
+          --exclude guardian-server-bench-loadgen
+          --exclude guardian-demo
+          --exclude guardian-rust-example
+
+      - name: Run tests with coverage (main)
+        if: github.event_name == 'push'
         run: >-
           cargo llvm-cov nextest
           --workspace
@@ -89,6 +105,7 @@ jobs:
           --doc
 
       - name: Upload coverage to Codecov
+        if: github.event_name == 'push'
         uses: codecov/codecov-action@0fb7174895f61a3b6b78fc075e0cd60383518dac # v5.5.5
         with:
           files: ./lcov.info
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -504,7 +467,11 @@ jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
-    if: github.repository_owner == 'OpenZeppelin'
+    # Main-only: this job proves the reproducible release profile (LTO,
+    # codegen-units=1) compiles with all features — ~5m per run that PR
+    # feedback does not need. A skipped job still reports a check run, so
+    # required-check rules stay satisfied on PRs.
+    if: github.repository_owner == 'OpenZeppelin' && github.event_name == 'push'
```
zeljkoX: Best to leave it per PR as well to discover build/link/compilation features before it merges to main.

zeljkoX (approved):
LGTM

Thanks for improving!

Added one comment.

PR #429, opened 2026-08-24
feat(sdk): historical public-note backfill by tag recovery primitive (#416)

### LINKED ISSUE #416: SDK recovery primitive: historical public-note backfill by tag

haseebrabbani:
Closes #416 (sub-issue of #357). Stacked on #423 and #424 — both branches are merged into this one because the backfill reuses their recovery modules, so the diff shows all three until they land in `357-recovery-primitives` and then collapses to the backfill alone. Review after those merge, or review just the two non-merge commits (`dab3757d`, `d28f0b76`).

Public notes addressed to an account are on chain, but normal forward sync starts from the store's **global** cursor — in a shared dirty store the cursor may already be past the blocks holding a recovered account's notes, and a fresh store has no efficient path at all. This adds the third recovery primitive from the #357 plan: a tag-scoped historical scan that discovers those notes and imports them individually with their on-chain inclusion proofs, without ever touching the global sync height.

## What's included

- Rust: `MultisigClient::backfill_public_notes_by_tag(account_id, from_block, to_block)` (genesis→tip defaults); TS: `backfillPublicNotesByTag(midenClient, { accountId, midenRpcEndpoint, fromBlock?, toBlock?, rpc? })`.
- Shared `PublicBackfillReport { scannedFrom/To, discovered, skippedPrivate, outcomes, uncovered, retryable, reason? }`, reusing `NoteImportOutcome` with a new `backfill` source. Scan problems are reported (`uncovered` ranges with `retryable`/`reason`), never thrown — a partial scan can't abort a recovery flow. Throws only when the scan range can't be established (tip lookup failure, bad account id, non-u32 bounds, inverted range).
- The node's per-request pagination cap is handled by client-side range halving under a bounded request budget (spike #412 finding); TS classifies it via the pinned `rpc pagination error` WASM fragment, added to the drift-guard test.
- Duplicates and tag collisions are tolerated per the spec (tags are best-effort filters; colliding public notes import harmlessly and simply never become consumable). Private matches are counted as `skippedPrivate` — the chain holds no body for them; the transport drain (#414) and proposal import (#415) cover those.
- A proof-less `Expected` record (e.g. left by a proposal import that ran while the note was uncommitted) is upgraded in place with the freshly fetched proof rather than skipped as already-present — forward sync will never revisit its block, so skipping would strand the note forever.
- Extracts the store-dedup (details-commitment keying) and with-proof import helpers now shared between the proposal import and the backfill, completing the shared-helper extraction flagged in #424.
- Docs: `docs/MULTISIG_SDK.md` (both SDK sections + API tables) and both package READMEs, including the caveat that notes sent with unrelated custom tags are outside this scan's guarantee.

## Follow-up candidates (not filed)

- Batch-first import with per-note fallback (shared with #415's flow) to collapse N store transactions in the happy path.
- Targeted TS store filters (List + Consumed + Expected) instead of the full pre-scan in `collectExistingRecords`.
- Structured error kinds on WASM `RpcClient` rejections, so pagination detection can match structurally like the Rust side instead of by message fragment.


crates/miden-multisig-client/README.md
```diff
@@ -308,6 +308,74 @@ let filter = NoteFilter::by_faucet_min_amount(faucet, 5_000);
 let spendable = client.list_consumable_notes_filtered(filter).await?;
 ```
 
+### Recovering Notes From Pending Proposals
```
zeljkoX: Let's review clients README files after all edits from parent branch are there. Maybe we should group all section under one so it has better context when to use and for what.

If it is too much content maybe even to move from README file to separate file and just link with higher overview in README file.

packages/miden-multisig-client/src/proposalNoteImport.ts
```diff
@@ -0,0 +1,451 @@
+/**
```
zeljkoX, line 1: what about introducing new dir for files introduced with parent ticket so they are easily navigated and have proper context. Now they are in src and no context connects them.

crates/miden-multisig-client/src/client/proposal_import.rs
```diff
@@ -0,0 +1,673 @@
+//! Recovery primitives for MultisigClient (issue #415, sub-issue of #357).
```
zeljkoX, line 1: proposal_note_import.rs filename

follows ts filename and has better context

crates/miden-multisig-client/src/client/public_note_backfill.rs
```diff
@@ -0,0 +1,1022 @@
+//! Historical public-note backfill by tag (issue #416, sub-issue of #357).
```
zeljkoX, line 1: public_note_backfill.rs

Follows ts filename

packages/miden-multisig-client/README.md
```diff
@@ -265,6 +265,110 @@ service's retention. Senders may deliver private notes out-of-band without
...
+`consume_notes` proposals embed the serialized notes they consume, and the
+standalone `importNotesFromProposals` helper turns those embedded bytes back
+into store records: it fetches each note's on-chain inclusion proof and
+imports the note individually, so it works for private notes too.
+
+```typescript
+import { importNotesFromProposals } from '@openzeppelin/miden-multisig-client';
+
+const proposals = await multisig.syncProposals();
+const outcomes = await importNotesFromProposals(midenClient, proposals, {
```
zeljkoX: Client should use method in multisig. That way no need to define rpc. Same approach as in other PR

packages/miden-multisig-client/README.md
```diff
@@ -265,6 +265,110 @@ service's retention. Senders may deliver private notes out-of-band without
...
+with the number of matching notes, not the range length, so a full
+genesis-to-tip scan is fast on an ordinary account.
+
+```typescript
+import { backfillPublicNotesByTag } from '@openzeppelin/miden-multisig-client';
+
+const multisig = await client.load(accountId, signer);
+const report = await backfillPublicNotesByTag(midenClient, {
+  accountId: multisig.accountId,
+  midenRpcEndpoint: 'https://rpc.testnet.miden.io',
```
zeljkoX: +1

packages/miden-multisig-client/README.md
```diff
@@ -265,6 +265,110 @@ service's retention. Senders may deliver private notes out-of-band without
...
```
zeljkoX: +1

docs/MULTISIG_SDK.md
```diff
@@ -627,6 +627,99 @@ transport, and relayed blobs are pruned after the retention window. Notes
...
+never needs to hold the note body.
+
+```typescript
+import { importNotesFromProposals } from '@openzeppelin/miden-multisig-client';
+
+const multisig = await client.load(accountId, signer);
+const proposals = await multisig.syncProposals();
+
+const outcomes = await importNotesFromProposals(midenClient, proposals, {
+  midenRpcEndpoint: 'https://rpc.testnet.miden.io',
```
zeljkoX: +1

zeljkoX (commented):
Thanks. Looks good.
Added minor comments.  Let's apply them before moving on.

PR #435, opened 2026-08-27
feat(sdk): note-recovery primitives — transport drain, proposal import, public backfill (#357)

### LINKED ISSUE #414: SDK recovery primitive: private-note transport backlog drain

### LINKED ISSUE #415: SDK recovery primitive: proposal-embedded note import

### LINKED ISSUE #416: SDK recovery primitive: historical public-note backfill by tag

haseebrabbani:
Closes #414, closes #415, closes #416. Advances #357 — together with the transaction-history endpoint (#413, merged via #420) this completes the recovery primitives from the #357 plan.

## Problem

After key-based recovery the local Miden store starts empty, and in a store shared with other accounts the global cursors may already be past the blocks and transport batches holding a recovered account's notes. Normal forward sync never revisits either, so pending notes are silently lost — the "hours of sync, still missing notes" scenario #357 describes. Spike #412 validated the recovery approach for each note class on a live network before implementation.

## The three primitives (both SDKs, semantic parity)

- **Private-note transport backlog drain (#414, PR #423)** — Rust `MultisigClient::drain_private_note_backlog()`, TS `drainPrivateNoteBacklog(midenClient)`. Rescans the full transport backlog for every tracked tag regardless of the stored cursor, without ever regressing it. Transport problems are reported (`completed` / `unavailable` / `failed` + retryability), never thrown, so a transport outage can't abort a recovery flow. Bounded by transport retention — a best-effort rescan, not a backup.
- **Proposal-embedded note import (#415, PR #424)** — Rust `MultisigClient::import_notes_from_proposals(&proposals)`, TS `multisig.importNotesFromProposals()`. Rebuilds importable notes from the bytes v2 consume-notes proposals embed (#229) plus a node-fetched inclusion proof, so it works for private notes too. Per-note outcomes (`imported` / `already-present` / `already-consumed` / `not-committed` / `invalid` / `failed`), dedup across proposals, and no per-note problem aborts the batch.
- **Historical public-note backfill by tag (#416, PR #429)** — Rust `MultisigClient::backfill_public_notes_by_tag(account_id, from, to)`, TS `multisig.backfillPublicNotesByTag()`. Tag-scoped historical scan (genesis→tip by default) that imports discovered public notes with their on-chain inclusion proofs, never touching the global sync height. Cost scales with matches, not range length (#412: genesis→tip over ~1.7M blocks in seconds). Handles the node's pagination cap by client-side range splitting, reports unscannable ranges instead of failing, and — like normal sync — screens every new discovery for relevance before import so tag collisions can't pollute the store.

## Cross-cutting design

- Shared per-note outcome model (`NoteImportOutcome` with a `source` discriminator) and shared store-dedup/import/consumed-reclassification helpers between the proposal import and the backfill.
- Store lookups key on details commitments (Rust) / id+details keys (TS) because metadata-less records (expected details imports, chain-consumed history) expose no note ID and would re-import forever.
- Report-don't-throw error philosophy throughout: recovery orchestration decides what to retry from structured reports; only environment failures (broken store, unresolvable scan range) surface as errors.

## Known limitations

- Transport recovery is bounded by the relay's retention window; out-of-band private sends are only recoverable from their sender.
- Notes sent with custom (non-standard) tags are outside the backfill's guarantee.
- The TS backfill screens statically against the well-known P2ID/P2IDE scripts because the WASM surface does not expose the execution-based `NoteScreener` (custom-script notes are conservatively skipped; upstream exposure would close the gap with Rust).


crates/miden-multisig-client/README.md
```diff
@@ -278,6 +278,140 @@ let filter = NoteFilter::by_faucet_min_amount(faucet, 5_000);
 let spendable = client.list_consumable_notes_filtered(filter).await?;
 ```
 
+### Recovering Notes After Device Loss
```
zeljkoX: nit: Maybe to move down this section a bit, for example after delta history, given it is not something executed often.

crates/miden-multisig-client/README.md
```diff
@@ -278,6 +278,140 @@ let filter = NoteFilter::by_faucet_min_amount(faucet, 5_000);
...
+client.pull_account(account_id).await?;
+
+let report = client.recover_notes(None).await?;
+println!("recovered {} notes", report.imported);
+for problem in &report.problems {
+    println!("step {} did not run: {}", problem.step, problem.reason);
+}
+```
+
+Pass `NoteRecoveryOptions` to choose strategies, bound the backfill's block
```
zeljkoX: Maybe to add small example with defined `NoteRecoveryOptions` for better visibility how to define strategies.

docs/MULTISIG_SDK.md
```diff
@@ -584,6 +584,179 @@ account executed elsewhere is not included. Output notes whose full
 details are not in the stored summary (e.g. private notes carried as
 partial notes) appear with `tag: 'custom'` and no recipient.
 
+### Recovering Notes After Device Loss
```
zeljkoX, line 557: nit: Move `Recovering Accounts By Key` this section just before `API Reference` section. It is natural that they stick together(now Delta History is in between). 
Also given it is not standard action best to put it after more used methods.

docs/MULTISIG_SDK.md
```diff
@@ -584,6 +584,179 @@ account executed elsewhere is not included. Output notes whose full
...
+  // A strategy that could not run at all lands here; the flow continues
+  // with the remaining strategies either way.
+  console.warn(`step ${problem.step} did not run: ${problem.reason}`);
+}
+if (report.retryable) {
+  // The flow is idempotent — rerun it to plausibly recover more.
+}
+```
+
+Pass `RecoverNotesOptions` to choose strategies (`transportDrain`,
```
zeljkoX: same as in rust readme, maybe small example of passing strategies

docs/MULTISIG_SDK.md
```diff
@@ -584,6 +584,179 @@ account executed elsewhere is not included. Output notes whose full
...
+`proposalImport`, `publicBackfill`), bound the backfill's block range
+(`fromBlock`/`toBlock`), or skip the final sync (`syncAfter: false`). The
+combined `NoteRecoveryReport` carries each strategy's own report
+(`transport`, `proposalImport`, `backfill`) untouched. The three
+strategies are also exposed individually, below.
+
+### Draining the Private-Note Transport Backlog
```
zeljkoX: Should this section be under `### Recovering Notes After Device Loss`. Now they are on same level. Same for bottom ones.

packages/miden-multisig-client/README.md
```diff
@@ -233,6 +233,136 @@ several accounts, and the helper surfaces all of them rather than silently
 picking one. The returned list is empty (not an error) when no account
 authorizes the queried commitment.
 
+### Recovering Notes After Device Loss
```
zeljkoX: nit: Maybe to move ### Recover An Account By Key and ### Recovering Notes After Device Loss
 toward bottom of document after more used methods given that these are not standard ones

crates/miden-multisig-client/src/client/note_recovery.rs
```diff
@@ -0,0 +1,417 @@
...
+    /// backfill range is inverted.
+    ///
+    /// # Example
+    ///
+    /// ```ignore
+    /// client.pull_account(account_id).await?;
+    /// let report = client.recover_notes(None).await?;
+    /// println!("recovered {} notes", report.imported);
+    /// ```
+    pub async fn recover_notes(
```
zeljkoX, line 162: If multiple strategies are selected should we run them in parallel?

crates/miden-multisig-client/README.md
```diff
@@ -278,6 +278,140 @@ let filter = NoteFilter::by_faucet_min_amount(faucet, 5_000);
...
+the cursor past private notes addressed to the newly recovered account.
+`drain_private_note_backlog` rescans the full transport backlog for every
+tracked note tag, regardless of the stored cursor. Run it after
+`pull_account`: the drain only sees tags tracked in the store, and
+`pull_account` inserting the account is what tracks its tag.
+
+```rust
+client.pull_account(account_id).await?;
+
+let report = client.drain_private_note_backlog().await?;
```
zeljkoX: When looking now, after we have higher level API, I am thinking that we should not expose internal methods for start given we have higher level ones `recover_notes`/`recoverNotes`. Keeping strategies implementation public increases the API and documentation surface and lets users accidentally skip required context or the verifying sync.

WDYT?

crates/miden-multisig-client/README.md
zeljkoX: Given that some of these could take time parallelization between selected strategies would be beneficial. However I am not sure could it introduce errors.

zeljkoX (commented):
LGTM

Added some nits

crates/miden-multisig-client/README.md
haseebrabbani: agreed, fixed

crates/miden-multisig-client/src/client/note_recovery.rs
haseebrabbani: Left sequential on purpose in both SDKs: Rust's `&mut self` means the strategies can't overlap anyway, and the TS WASM client runs calls one at a time, so parallelizing would only overlap network waits while breaking the report. The backfill scan dominates the runtime and is already a single streamed request, so the win would be small; happy to revisit if profiling shows otherwise.

zeljkoX (approved):
LGTM

PR #436, opened 2026-08-27
feat(multisig-client): offline switch-guardian proposal creation (#433)

### LINKED ISSUE #433: TS client: no offline proposal creation — `createSwitchGuardianProposal` pushes to the current GUARDIAN before caching anything (Rust has `create_proposal_offline`)
## Problem

The offline switch path exists so an account can leave an unreachable operator: `executeProposal` for `switch_guardian` uses the locally cached tx summary, skips the GUARDIAN ack, treats the old-operator push as best-effort, and registers on the new operator from local state. The Rust client exposes the full trio for this (`create_proposal_offline` / `sign_imported_proposal` / `execute_imported_proposal`, `crates/miden-multisig-client/src/client/offline.rs`), documented as "Only `SwitchGuardian` transactions can be executed fully offline."

The TS package has import/sign/execute (`importProposal`, `signProposalOffline`) but no creation counterpart. `createSwitchGuardianProposal` funnels into `createProposal`, which calls `pushDeltaProposal` against the account's current GUARDIAN **before** caching the proposal locally (`packages/miden-multisig-client/src/multisig.ts`). Against a dead operator the call throws and the locally built summary/metadata are lost — so the one proposal type designed to work without the current operator cannot be created without it.

A TS consumer can work around this by hand-assembling the `ExportedProposal` JSON for `importProposal` from the exported primitives (`buildUpdateGuardianTransactionRequest`, `executeForSummary`, `chainAnchorToBase64`), but the commitment has to be derived via the SDK (`TransactionSummary.toCommitment()`) because `computeCommitmentFromTxSummary` is not exported. This asymmetry is a direct contributor to 0xMiden/wallet#782, where the wallet routed the switch through the outgoing operator's HTTP API and an account got stuck on a network-dead operator.

## Proposal

- Add `createSwitchGuardianProposalOffline(newGuardianEndpoint, newGuardianPubkey, options)` (or a general `createProposalOffline` gated to switch, mirroring Rust): build the request and summary locally, verify the new endpoint's `/pubkey` commitment, include the proposer's signature, cache the proposal locally, and return the `ExportedProposal` — no push to the current operator.
- Export `computeCommitmentFromTxSummary` (or document the SDK-based derivation) so hand-rolled flows are at least possible without guesswork.
- Unit tests mirroring the Rust offline tests: create → sign (second signer) → execute with the current-operator client pointed at an unreachable endpoint.

haseebrabbani:
Closes #433.

Adds `createSwitchGuardianProposalOffline(newGuardianEndpoint, newGuardianPubkey, { nonce }?)` to the TS client — the creation counterpart to the existing `importProposal` / `signProposalOffline` / `executeProposal` trio, mirroring the Rust `create_proposal_offline`. Nothing is pushed to the current GUARDIAN, so an account can leave an unreachable operator (the 0xMiden/wallet#782 scenario): the method syncs with the Miden node (Rust `sync_network_only` parity), verifies the new endpoint's `/pubkey` commitment, builds and signs the summary locally, caches the proposal, and returns the `ExportedProposal` (proposer signature included) for side-channel cosigning.

Design notes:
- The verify/build/metadata step is extracted into a private `buildSwitchGuardianSummary` shared with `createSwitchGuardianProposal`, so the online and offline proposals for the same operation cannot drift.
- Caching and proposer signing reuse `importProposal` + `signProposalOffline` end to end rather than re-implementing them, so the local copy carries exactly the guarantees a cosigner-imported one does.
- `computeCommitmentFromTxSummary` is now exported (second ask in the issue) and returns normalized hex, directly comparable to `ExportedProposal.commitment` / `Proposal.id`.
- Only switch-guardian proposals can be created offline — every other type needs a GUARDIAN ack at execution (same restriction as Rust, moved to compile time by the per-type method).

### DIFF
--- a/packages/miden-multisig-client/src/multisig.ts
+++ b/packages/miden-multisig-client/src/multisig.ts
@@ -925,6 +925,27 @@ export class Multisig {
     options: CreateProposalOptions = {},
   ): Promise<Proposal> {
     const proposalNonce = resolveProposalNonce('createSwitchGuardianProposal', options);
+    const { summaryBase64, metadata } = await this.buildSwitchGuardianSummary(
+      newGuardianEndpoint,
+      newGuardianPubkey,
+    );
+
+    // SwitchGuardian is a regular delta proposal; push it to GUARDIAN so
+    // sign/execute (which fetch from GUARDIAN) can find it. To leave an
+    // unreachable GUARDIAN, use createSwitchGuardianProposalOffline instead.
+    return this.createProposal(proposalNonce, summaryBase64, metadata);
+  }
+
+  /**
+   * Shared build step for both switch-GUARDIAN creation paths: verify the new
+   * endpoint's `/pubkey` commitment, then execute the update-guardian request
+   * for its summary and metadata. Kept in one place so the online and offline
+   * proposals for the same operation can never drift apart.
+   */
+  private async buildSwitchGuardianSummary(
+    newGuardianEndpoint: string,
+    newGuardianPubkey: string,
+  ): Promise<{ summaryBase64: string; metadata: ProposalMetadata }> {
     const webClient = await this.getRawClient();
     await this.verifyGuardianEndpointCommitment(newGuardianEndpoint, newGuardianPubkey);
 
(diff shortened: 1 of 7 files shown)


packages/miden-multisig-client/src/multisig.ts
```diff
@@ -949,9 +970,68 @@ export class Multisig {
...
+    options: CreateProposalOptions = {},
+  ): Promise<ExportedProposal> {
+    const proposalNonce = resolveProposalNonce('createSwitchGuardianProposalOffline', options);
+
+    // Sync with the Miden node before building (mirrors the Rust
+    // `sync_network_only`): with no GUARDIAN push to reject a stale delta at
+    // creation, a summary built from stale local state would only fail at
+    // execution — after the whole side-channel cosigning ceremony.
+    const webClient = await this.getRawClient();
+    await retryRpcRead(() => webClient.syncState(), this.rpcConfig);
```
zeljkoX: Rust’s `sync_network_only()` also refreshes the cached account after syncing. Here, `webClient.syncState()` updates the WASM store but leaves `this.account`, `this.threshold`, `this.signerCommitments`, and `this.procedureThresholds` at their previous values.

That means the summary can be built against the latest account state while readiness and signature validation use stale configuration. For example, after the on-chain threshold changes from 1 to 2, this client could incorrectly consider the proposal ready with one signature and fail only during submission.

Please reload the account from the store and call `refreshConfigFromAccount()` after this sync, ideally through a private `syncNetworkOnly()` helper shared with the post-switch sync in `executeProposal`.

zeljkoX (approved):
LGTM

Added one minor commment.

PR #440, opened 2026-08-28
feat(multisig): commit fee conversion info and pay the fee on guarded accounts

WiktorStarczewski:
> **Read this as four PRs.** The same content is split into a stack of four readable diffs, in dependency order. This PR is unchanged and stays open as the whole-change view and the home of the existing review threads.
>
> 1. [Link the guarded auth component the way both SDKs do](https://github.com/OpenZeppelin/guardian/pull/450) — 20 files, +632/-97 — **open here, reviewable now**
> 2. [Pay the guarded-account fee from the Rust client](https://github.com/WiktorStarczewski/guardian/pull/2) — 25 files, +1579/-65
> 3. [Commit fee conversion info in the TypeScript request builders](https://github.com/WiktorStarczewski/guardian/pull/3) — 23 files, +1226/-39
> 4. [Commit the anchored block's fee faucet on every proposal](https://github.com/WiktorStarczewski/guardian/pull/4) — 6 files, +2023/-99
> 
> PRs 2-4 sit in the fork because a PR's base branch must live in the base repo and I have no
> push access here. To surface them: push `split/pr1-linkage`, `split/pr2-rust-fee` and
> `split/pr3-ts-fee-auth` to this repo, then 2-4 can be reopened against them and this PR
> retargeted onto `split/pr3-ts-fee-auth`, shrinking it to diff 4 with its review threads intact.
>
> The four together are byte-identical to this branch. Each was verified to build, lint and test on its own base.
>
> They are opened inside the fork because the stack needs its base branches to live in the base repo, and I have no push access to `OpenZeppelin/guardian`. **To move the stack here, a maintainer needs to push `split/pr1-linkage`, `split/pr2-rust-fee` and `split/pr3-ts-fee-auth` to this repo** — they exist on `WiktorStarczewski/guardian` and are ready to fetch. Once they are here the four PRs can be reopened fork→origin and this one retargeted to `split/pr3-ts-fee-auth`, at which point it shrinks to exactly diff 4 with its review threads intact.

Guardian-custodied accounts could not transact on a chain that charges a transaction fee. This makes them pay it.

- `transaction/feeAuth.ts` builds the auth arg protocol 0.16 requires: the commitment `hash(CONVERSION_INFO || SALT)`, with its preimage in the advice map.
- `guarded_multisig.masm` pays the fee, mirroring the upstream protocol change.
- The fee faucet is read from a block header and threaded through **all seven** typed proposal creators and the rebuild paths.
- `resolveAuthArg` / `nativeConversionInfo` / `feeAuthArg` are exported, so `createCustomProposal` callers can make a custom proposal payable — they had no supported way to before.
- `prepareProposalExecution` recovers the salt from `metadata.saltHex` rather than the summary, because the summary's user-params word is now the commitment.

<details>
<summary>Why the create paths now commit conversion info (invariant change)</summary>

The typed create paths used to pass a **bare salt**, so the Rust SDK could rebuild a proposal from `salt_hex`. Under protocol#3765 the guarded auth calls `fee::pay_fee` **before** the summary is built, so a bare arg aborts the create-time execution with `ERR_FEE_CONVERSION_INFO_MISSING` — a structural proposal could not be created at all on a fee-charging chain.

Cross-SDK reconstruction is preserved by committing **native** conversion info: `getFeeFaucetId()` returns the chain's own fee faucet, so the arg is `hash(native_conversion_info(chain fee faucet) || salt)`, rederivable from `salt_hex` plus the reference block exactly as `fee::native_conversion_info` does. The invariant weakens deliberately from "the auth arg **is** the salt" to "the auth arg is **derivable from** the salt", and `multisig.test.ts` asserts the new form.

This also removes an inconsistency: `createP2idProposal` and `createConsumeNotesProposal` already committed conversion info, while the five structural creators were forbidden from it.
</details>

<details>
<summary>The ordering asymmetry, and the Poseidon2 equivalence</summary>

The advice-map value is `SALT ++ CONVERSION_INFO`, but the commitment hashes `CONVERSION_INFO ++ SALT`. This matches `commit_fee_conversion_info` on the Rust side.

The MASM uses `poseidon2::merge`; JS exposes only `Poseidon2.hashElements`. They agree for exactly eight field elements — `merge` seeds a zero capacity, and `hash_elements` sets `capacity[0] = len % RATE_WIDTH`, which is zero at eight. The length is asserted rather than assumed, because the equivalence holds only there.
</details>

**Depends on [0xMiden/protocol#3765](https://github.com/0xMiden/protocol/pull/3765)**, which makes the guarded component pay the fee.

The dashboard's fee-note exclusion started here and now ships separately as #449: it is server-side, shares no code with this, and builds green against published protocol crates.

Verified end to end against a localnet at `verification_base_fee = 10000`, with the node and SDK built from #3765: the consuming wallet's guardian E2E suite goes from every send failing to 49 specs passing, covering sends, claims, device-key rotation and guardian switch.

**Reviewers:** the invariant change in the first `<details>` is the part worth your time — whether "derivable from the salt" is genuinely enough for the Rust SDK to rebuild, since that is what replaces the bare-salt guarantee.


haseebrabbani:
@WiktorStarczewski can you address the merge conflicts before we review?

WiktorStarczewski:
Split into four stacked PRs, all open and ready for review:

| | | |
|---|---|---|
| 1 | Build guarded accounts from the standard auth component | #450 |
| 2 | Pay the guarded-account fee from the Rust client | #451 |
| 3 | Commit fee conversion info in the TypeScript request builders | #452 |
| 4 | Commit the anchored block's fee faucet on every proposal | #453 |

Each targets its parent, so the diff shown is that step alone. #450 is green on all 14 checks.

One thing changed substantively since this PR: the first step now builds guarded accounts from the SDK's standard `AuthGuardedMultisig` component instead of a vendored copy of its MASM. A locally compiled copy links the standards package dynamically and roots differently, so `AccountComponentInterface::from_procedures` couldn't classify the account and the client never attached fee conversion info — which is the thing this whole stack exists to make work.

Closing in favour of the split. The review threads here stay for history.

Issue #418, opened 2026-08-19
Deferred: account creation-block anchor for bounded recovery scans


haseebrabbani:
**Status: deferred optimization.** Filed to record the design decision and the benchmark rationale from spike #412 (part of the #357 recovery work). Not blocking anything.

## Proposal

Record the account's creation block height at `/configure` time and return it from `getState` as an additive optional field. A recovering wallet then scans `[creation_block, tip]` instead of `[genesis, tip]` for historical note discovery, and a wallet-backup watermark gains a trustworthy lower bound.

## Why it's deferred

Spike #412 measured genesis→tip tag-scoped `SyncNotes` on testnet (~1.61M blocks): 0.34 s with zero matches, ~2 s at ~3k matching blocks, 44.5 s at 135k matching blocks. Cost scales with **matches, not range length**, because the node's pagination skips non-matching spans — so a creation-block lower bound saves little at current network scale, and the #357 public-backfill primitive ships with genesis as its v1 lower bound.

## Revisit triggers

Reopen prioritization if any of these change the constants:
- Mainnet chain length / block density / note volume makes genesis scans materially slower.
- The miden-client 1000-page pagination safety cap starts being hit in practice on full-range scans.
- Wallet-backup export preflights (which need a proven completeness lower bound) become a frequent operation.

## Sketch (when picked up)

- Additive column on account metadata + capture at `/configure` (or first canonical delta as fallback for existing accounts).
- Additive optional field in `getState` (HTTP + gRPC) — no breaking contract change.
- Recovery/backfill SDK paths use it as `fromBlock` when present, genesis otherwise (already their fallback).
- Consider also recording the landing block per canonical delta while in the area (bounds consumption-status reconstruction).

PR #441, opened 2026-08-28
feat(sdk): preserve proposal-embedded notes on the switch-guardian path (#417)

### LINKED ISSUE #417: Run note-recovery primitives on the switch-guardian path
Sub-issue of #357. Follow-up: depends on the three SDK recovery primitives (transport drain, proposal-note import, public backfill).

## Context

Switching Guardian operators loses the same local note context as device-loss recovery: the switch path performs a chain sync and re-registers state on the new Guardian, but carries no note or history context. Any note-recovery capability built for #357 should be exercised (or explicitly scoped out) on the switch path too.

## Scope

- Invoke the transport backlog drain and public-note backfill on the switch-guardian flow where they apply; include proposal-embedded note import if pending proposals survive the switch.
- If any primitive is deliberately skipped on this path, document why.

## Acceptance criteria

- [ ] Switch flow either runs the applicable recovery primitives or documents the decision not to.
- [ ] A test covering note continuity across a guardian switch (note received before switch remains consumable after).

Low urgency; small once the primitives exist.

haseebrabbani:
Closes #417 (last active sub-issue of #357; #418 remains deferred).

## What

Pending proposals do not survive a guardian switch — the new GUARDIAN is registered with bare account state — so the notes embedded in pending consume-notes proposals are the one recovery source a later device-loss `recover_notes` / `recoverNotes` can no longer reach once the old GUARDIAN is left behind. Both SDKs now run the proposal-embedded note import (#415 primitive) on the online switch path, against the old GUARDIAN, before the switch transaction executes.

## Why only the proposal import

Per the issue's "document what is skipped" criterion: the transport drain and public backfill are deliberately skipped on the switch path. A switch executes against an intact local store (chain cursor, transport cursor, and note records all persist), and both the note transport and the node are configured independently of the GUARDIAN — a switch loses nothing those primitives rescan. They remain device-loss recovery tools and work unchanged against the new GUARDIAN. The offline switch flow also skips the import (it exists to avoid contacting the GUARDIAN); its docs now say so and point at the manual call.

## Design notes

- The import runs pre-execution because per-proposal summary-binding re-verification only reproduces while the account has not advanced past the state the proposals were built on; it also runs before the best-effort delta push, so the pending listing is read before anything switch-related lands on the old GUARDIAN.
- Best-effort and bounded: failures and per-note `failed`/`invalid` outcomes are warned (this is the last moment the notes are reachable, so "retryable" cannot help), and the whole flow runs under a 30s timeout (`PRE_SWITCH_IMPORT_TIMEOUT` / `PRE_SWITCH_IMPORT_TIMEOUT_MS`) — neither guardian client applies request deadlines, so a hung old GUARDIAN must not stall the switch.
- The switch slice is a named, exhaustively-listed preset in both SDKs (`NoteRecoveryOptions::for_guardian_switch()`, exported `GUARDIAN_SWITCH_RECOVERY_OPTIONS`), so adding a recovery strategy forces an explicit decision about the switch path. `preserve_pre_switch_proposal_notes` / `preservePreSwitchProposalNotes` are public, and the manual repoint seams (`set_guardian_endpoint`, `setGuardianClient`) document calling them first.
- Drive-by: the TS proposal listing now checks the stale-nonce filter before the per-proposal binding re-execution, matching the Rust listing's order.

## Known boundaries

- Proposals superseded at an earlier nonce (lost a same-nonce race, embedded note still unconsumed) are filtered by the listing and not imported — a pre-existing gap shared with device-loss recovery.
- The listing binding-verifies every pending proposal (a VM re-execution each) although the import only consumes consume-notes v2; a type pre-filter in the shared listing would take that off the switch critical path.


zeljkoX (commented):
Automatic preservation on the online SwitchGuardian execution path is the right default: this is the last point where pending proposals on the old Guardian remain available. The public/manual methods are still useful for direct client repointing and optional preparation before offline execution.

I found the following issues to address before merging:

**1. Missing continuity coverage**

Issue #417 requires proving that a note received before the switch remains consumable afterward. The Rust test only exercises the recovery-options slice against an unreachable Guardian, while the TypeScript test mocks `importNotesFromProposals`. Neither verifies post-switch consumability.

Recommendation: add one continuity test on at least one SDK (ideally both): a distinct v2 `consume_notes` proposal → switch → note still in the store and consumable. Do not reuse the shared `AQID` summary.

**2. TypeScript timeout leaves recovery running**

Rust’s timeout cancels the recovery future and returns `Option<NoteRecoveryReport>`. TypeScript returns `void` and uses `Promise.race`, so the recovery promise stays active after timeout. That leftover still uses the same `Multisig` instance — later `this.guardian` reads can hit the new client after `setGuardianClient()`, and import/store access can interleave with `submitAt()` on the shared WASM client.

Recommendation: `preservePreSwitchProposalNotes` should return `Promise<NoteRecoveryReport | undefined>` (mirror Rust). Abort in-flight listing/import on timeout instead of `flow.catch(() => {})`. Do not start `submitAt` until that work is dead.

**3. Aliased TypeScript fixture**

The pending delta reuses the switch proposal’s `AQID` summary, so it inherits cached switch metadata. A real `importNotesFromProposals` invocation would skip it rather than exercise a `consume_notes` proposal.

Recommendation: use a distinct summary with genuine `consume_notes` metadata and note bytes, and assert the import received `proposalType: 'consume_notes'`.

**4. Missing Rust execute-path wiring test**

There is no Rust test proving that `execute_proposal()` invokes preservation specifically for SwitchGuardian. Removing that call would currently go unnoticed.

Recommendation: add an `execute_proposal(SwitchGuardian)` wiring test that the listing/import ran, matching the TypeScript order test (`import → execute → register`).

**5. Public API documentation**

Please document `preserve_pre_switch_proposal_notes`, `NoteRecoveryOptions::for_guardian_switch()`, `preservePreSwitchProposalNotes`, and `GUARDIAN_SWITCH_RECOVERY_OPTIONS` in both package READMEs and `docs/MULTISIG_SDK.md`. The docs should clarify:

- automatic for online switch execution
- explicit before direct client repointing
- deliberately skipped during offline execution

No critical issue found, but the timeout lifecycle and missing end-to-end continuity coverage should be resolved before merge.

zeljkoX (approved):
LGTM

Let's revisit inline comments. They seem big and repetitive. Keeping them short and moving rationale to separate /docs document to document recovery mechanism makes sense to me.

PR #450, opened 2026-09-03
fix(multisig): build guarded accounts from the standard auth component

WiktorStarczewski:
Guarded accounts are built from the SDK's standard `AuthGuardedMultisig` component instead of a vendored copy of its MASM, so the client can classify them and attach fee conversion info — without which every guarded transaction fails on a fee-charging chain.

- Compiling an equivalent locally links `StandardsLib` dynamically and yields a different `auth_tx` root. `AccountComponentInterface::from_procedures` matches on that root, and an unmatched account falls through to `FeeAuth::Ignored` — no conversion info, no fee paid.
- Both builders now call upstream: Rust `AuthGuardedMultisig::new`, TypeScript `createAuthGuardedMultisig`. The root agrees by construction rather than by pin.
- Removes the vendored MASM, its generator, and the generated modules; `auth_tx` is repointed at upstream's root.
- `procedure_roots_match_upstream_component` now covers `auth_tx` too. It was excluded only because the vendored copy rooted differently, so deleting that copy strengthens the guard for free.
- Cross-SDK parity constants regenerated. Storage commitment and slot count are unchanged — only the code commitment moved, which is what a component swap should do.

Requires `@miden-sdk/miden-sdk` 0.16.0-rc.7, which adds the `createAuthGuardedMultisig` export. Its `AuthScheme` discriminants are inlined rather than imported — the package root exports two different things under that name and the string-valued one wins, so importing it yields `undefined`. Tracked as [0xMiden/web-sdk#365](https://github.com/0xMiden/web-sdk/issues/365).

**Reviewers:** this reverses the linkage approach in the previous revision. Making the two roots agree dynamically left both of them unclassifiable, which is the defect that actually mattered — that reasoning is the part worth checking.

---

<details>
<summary>Why not keep the dynamic linkage</summary>

The previous revision compiled the vendored MASM through `CodeBuilder` on both sides, so Rust and the browser produced one consistent root. That fixed the symptom the suite showed — roots disagreeing — without fixing the cause: the shared root still belonged to no known component, so `from_procedures` could not classify it either way.

Classification is what gates fee support. `FeeAuth::of` returns `Ignored` for an unrecognised auth component, the client then declines to attach fee conversion info, and the transaction cannot pay its fee from the account vault. A caller-supplied salt does not rescue it — that path returns `FeeConversionInfoUnsupported`.

Building from upstream makes the root correct by construction and removes the vendored copy that could drift from it.
</details>

<details>
<summary>Verification</summary>

| | |
|---|---|
| Rust `--lib` | 1228 passed / 1 failed |
| `cargo clippy --workspace --all-targets -D warnings` | clean |
| `cargo fmt --check` | clean |
| TypeScript (vitest) | 559 passed / 559, 47 files |
| Browser determinism gate | 2 passed |
| `account-roundtrip` | passes — this was the branch's failing test |

The one Rust failure is `client::note_recovery::tests::recover_notes_continues_past_failing_steps`, a local-environment artifact rather than a defect: it asserts the direct node channel is unreachable, so it fails on any machine running a local `miden-node` (its `scanned_to` tracks the live chain tip). Green in CI on `main`, and untouched by this branch.

The toolchain bump to 1.98 surfaces one new clippy lint in `examples/demo`; fixed here since the bump is what exposes it.

The browser determinism gate passes against rc.7 in a real browser, so the regenerated parity constants are confirmed from both sides rather than copied from the Rust run.

</details>

---

<sub>

**The stack** — each PR targets its parent, so the diff shown is that step alone.

| | | |
|---|---|---|
| 1 | Build guarded accounts from the standard component | [#450](https://github.com/OpenZeppelin/guardian/pull/450) |
| 2 | Pay the guarded-account fee from the Rust client | [#451](https://github.com/OpenZeppelin/guardian/pull/451) |
| 3 | Commit fee conversion info in the TypeScript request builders | [#452](https://github.com/OpenZeppelin/guardian/pull/452) |
| 4 | Commit the anchored block's fee faucet on every proposal | [#453](https://github.com/OpenZeppelin/guardian/pull/453) |

Each is retargeted to `main` as its parent merges. Supersedes [#440](https://github.com/OpenZeppelin/guardian/pull/440), which carried all four as a single +5457 diff; its review threads stay there for history.

</sub>


.github/workflows/ci.yml
```diff
@@ -40,7 +40,7 @@ jobs:
       - name: Install Rust
         uses: dtolnay/rust-toolchain@351f82a4dc29e4159746a068ed925da17341219f # 1.89.0
         with:
-          toolchain: 1.96.1
+          toolchain: 1.98.0
```
zeljkoX, line 43: let's bump toolchain in publish-crates.yml as well

crates/contracts/src/multisig_guardian.rs
```diff
@@ -4,19 +4,61 @@
...
+/// The source is the same; the *linkage* is not, and that is why this constant exists rather
+/// than a plain `AuthGuardedMultisig::code()` call. `auth_tx_guarded_multisig` is the only
+/// export that calls `miden::standards::fee`, so it is the only one whose root moves with how
+/// the standards package is linked: upstream's component manifest declares
+/// `miden-standards = { linkage = "static" }` and inlines the callee's MAST, while
+/// [`CodeBuilder`] links it dynamically and leaves an external reference. Building through
+/// [`CodeBuilder`] here is what makes a Rust-built account carry the same `auth_tx` root as a
+/// browser-built one — the root `miden_multisig_client::ProcedureName::AuthTx` pins, and the
+/// key the per-procedure threshold overrides map is keyed by.
+const GUARDED_MULTISIG_AUTH_MASM: &str = include_str!(
```
zeljkoX: `guarded_multisig.masm` is included in `crates/contracts/src/multisig_guardian.rs`, `crates/miden-multisig-client/examples/procedure_roots.rs`, and the `procedures.rs` tests, and each site recompiles it through `CodeBuilder`. 

Suggestion:

- make `GUARDED_MULTISIG_AUTH_MASM` `pub` in the contracts crate (it already owns the build)
- have the example and the `procedures.rs` tests call the existing `dynamically_linked_auth_code()` instead of compiling again; `miden-multisig-client` already depends on `miden-confidential-contracts`, so no new dependency
- only `bundled_auth_code_statically_linked()` needs the raw string, and it can import the public constant

zeljkoX (commented):
Thanks for working on this. 

Decision behind this PR looks good to me.

Do we know when web-sdk is going to be published? Should this PR wait for it?

WiktorStarczewski:
Thanks — both addressed in `7adc2ee`, and the web-sdk question has an answer now.

**`publish-crates.yml` toolchain** — bumped both blocks (validate + publish) to `1.98.0`. Worth noting it was the last stale pin in the repo, and the drift predates this PR: the hardcode matched `rust-toolchain.toml` when the file was written at `ca67dc0`, then #425/#427 moved the toolchain file without it. It has been inert rather than broken — a root `rust-toolchain.toml` overrides `rustup default`, so cargo was already running 1.98.0 and the pin only cost a redundant toolchain download. But it is a real trap: all four crates it publishes carry `rust-version.workspace = true`, so anything that stops the toolchain file applying (adding `RUSTUP_TOOLCHAIN`, moving the file) turns it into a hard `rustc 1.93.1 is not supported` failure on a `release: published` run, which is the worst place to find it.

Heads up that touching this file activates its `paths:` trigger, so #450 now runs a `cargo publish --dry-run` over all four crates that it was not running before. That is a useful early warning, but if it goes red it will be red on something new rather than something I broke.

**Sharing the MASM constant** — done, and you were right that three sites were compiling the same file. `GUARDED_MULTISIG_AUTH_MASM` is now `pub`, and both the example and the `procedures.rs` tests import it; one `include_str!` remains, in the contracts crate that owns it. I added a doc line recording that this path must stay the TypeScript package's MASM, since `generate-masm.mjs` vendors the same file into `auth.ts` and the cross-SDK parity checks anchor on it.

I did **not** take the last part — swapping `bundled_auth_code()` for `dynamically_linked_auth_code()`. Once the constant is shared those two are the same computation, so it only removes a three-line call, and it costs some failure-diagnosticity: today a drift reddens only the account test, which tells you production drifted while the MASM and the pin are fine. After the swap `auth_tx_root_matches_the_bundled_masm` becomes a strict subset of `rust_built_accounts_carry_the_pinned_auth_tx_root` and both go red together. Happy to take it if you would rather have the smaller surface.

`cargo check --all-targets` is clean and all 10 `procedures` tests pass, including the four root-parity ones.

**On waiting for web-sdk:** no need — `@miden-sdk/miden-sdk@0.16.0-rc.6` is publishing right now. It is the first release carrying the fee-conversion surface, built against the published `miden-client 0.16.0-rc.4` (protocol `0.16.0-rc.9`). That should also clear this PR's two red checks: they fail because the published rc.5 SDK embeds protocol rc.6 and cannot classify an rc.9 guarded account, so `from_procedures` finds zero auth components and `AccountInterface::new` traps. I have left the guardian-side `@miden-sdk/miden-sdk` bump to you rather than folding a version change into this PR.

WiktorStarczewski:
heads up: reversed the approach you reviewed.

dynamic linkage made the two roots agree, but the shared root still matched no known component — so `from_procedures` couldn't classify the account and it never got fee conversion info. that was the actual bug, the root mismatch was just the visible half.

now both builders use upstream's `AuthGuardedMultisig`, vendored masm + generator deleted. `account-roundtrip` passes, and the browser determinism gate confirms the regenerated parity constants from both sides. pinned to web-sdk rc.7 (published).

description is rewritten rather than patched, so probably worth a re-read over a diff. #451-453 are the rest of the stack, rebased onto this.

haseebrabbani:
mainly comment on docs and release handling:

- **Docs describe the reversed design.** `crates/contracts/README.md:7-12` and `docs/MULTISIG_SDK.md:1646-1653` still say the auth component is compiled from vendored MASM through `CodeBuilder` and cite a `GUARDED_MULTISIG_AUTH_MASM` symbol that no longer exists.
- **`docs/MIDEN_COMPATIBILITY.md` contradicts the tree.** The matrix row and the "mid-move" paragraph say the web SDK is stuck at rc.5 and the browser builder is held back, while every package.json here pins rc.7 and the browser gate is what this PR makes pass. `docs/MULTISIG_SDK.md:30` and the package README still install rc.5
- **Breaking release with no version bump.** The `auth_tx` root moved, so accounts built by the published 0.17.0-rc.2 now fail `assertPinnedContractVersion` and `UnsupportedContractVersion`. `MULTISIG_SDK.md:1673-1680` says a root-changing pin bump is a breaking release: bump the version, add a matrix row, and note it in release notes. The PR edits the rc.2 row in place, misdescribing a tagged, published release.
- **miden-client pin mismatch.** The rc.7 wasm embeds `miden-client-0.16.0-rc.4`; `Cargo.toml:88` keeps `=0.16.0-rc.3`. AGENTS.md §11 requires the Rust pins to match the wasm exactly, and rc.4 is on crates.io as of today.

docs/MIDEN_COMPATIBILITY.md
```diff
@@ -38,6 +38,17 @@ SDK's embedded `miden-standards` matches the Rust pin, so the CI parity gates ar
 what catch drift. See
 [`MULTISIG_SDK.md`](./MULTISIG_SDK.md#contract-version-pinning).
 
+**The 0.17.0-rc.2 row is mid-move.** The guarded-multisig auth component now pays the
+transaction fee, which needs `miden::standards::fee` procedures that first ship in
+`miden-standards` 0.16.0-rc.9 — so the Rust pin is there. The newest published
+`@miden-sdk/miden-sdk` is still 0.16.0-rc.5, which embeds an older `miden-standards`: it
```
zeljkoX: this is now outdated.

Cargo.toml
```diff
@@ -80,10 +80,10 @@ utoipa = { version = "5", features = ["axum_extras"] }
...
-miden-protocol = "=0.16.0-rc.6"
-miden-standards = "=0.16.0-rc.6"
-miden-tx = "=0.16.0-rc.6"
+miden-protocol = "=0.16.0-rc.9"
+miden-standards = "=0.16.0-rc.9"
+miden-tx = "=0.16.0-rc.9"
 miden-node-proto-build = { version = "=0.16.0-rc.3", default-features = false }
-miden-testing = "=0.16.0-rc.6"
+miden-testing = "=0.16.0-rc.9"
 miden-client = "=0.16.0-rc.3"
```
zeljkoX: Should we bump it to rc.4?

docs/MULTISIG_SDK.md
```diff
@@ -1741,6 +1643,15 @@ one:
 - **TypeScript**: the `@miden-sdk/miden-sdk` pin, whose bundled WASM embeds the
   matching upstream `miden-standards` guarded-multisig component
 
+Matching versions is necessary but not sufficient: the *linkage* has to match too.
```
zeljkoX: This needs to be updated to match new approach.

crates/contracts/README.md
```diff
@@ -3,8 +3,13 @@
 Builder facade for Guardian's custody accounts on [Miden](https://miden.xyz).
 
 Since the adoption of the audited upstream `AuthGuardedMultisig` component from
-[`miden-standards`](https://crates.io/crates/miden-standards), this crate no longer
-ships any MASM of its own. It provides:
+[`miden-standards`](https://crates.io/crates/miden-standards), this crate authors no
```
zeljkoX: Needs update after latest changes.

docs/MIDEN_COMPATIBILITY.md
```diff
@@ -20,7 +20,7 @@ elsewhere and link here:
 
 | Guardian | Miden protocol | `miden-protocol` / `miden-standards` | `miden-client` (Rust) | `@miden-sdk/miden-sdk` (npm) |
 |---|---|---|---|---|
-| 0.17.0-rc.2 | 0.16 (rc) | `=0.16.0-rc.6` | `=0.16.0-rc.3` | `0.16.0-rc.5` (exact) |
+| 0.17.0-rc.2 | 0.16 (rc) | `=0.16.0-rc.9` | `=0.16.0-rc.3` | `0.16.0-rc.5` (exact) |
```
zeljkoX: `=0.16.0-rc.9` | `0.16.0-rc.4` | `0.16.0-rc.7 ?

WiktorStarczewski:
All four were real. Fixed in `6eb1a40..0f57716`:

**Docs describing the reversed design** — `crates/contracts/README.md` and `MULTISIG_SDK.md` now say both SDKs take the component from `miden-standards` and compile nothing, so the roots are upstream's by construction. The `GUARDED_MULTISIG_AUTH_MASM` reference is gone; the symbol was deleted with the vendored MASM.

**`MIDEN_COMPATIBILITY.md` contradicting the tree** — install lines moved rc.5 → rc.7, and the "mid-move" paragraph is replaced. It described a gap rc.7 closed.

**Breaking release with no bump** — you were right that editing the rc.2 row in place misdescribed a tagged release. Restored it to what shipped (`rc.6` / `rc.3` / `rc.5`) and added a real `0.17.0-rc.3` row; version bumped across the workspace and all five npm packages. The compatibility doc now states plainly that rc.2 accounts fail `assertPinnedContractVersion` under rc.3 and that there is no in-place migration.

**miden-client pin** — confirmed against the shipped wasm rather than the tag: `strings … miden_client_web.wasm` gives `miden-client-0.16.0-rc.4`, protocol/standards/tx/testing all rc.9. Pinned to rc.4.

One thing worth flagging, since it went past what the comment described: rc.4 needs `miden-node-proto-build ^0.16.0-rc.5` and we pinned `=0.16.0-rc.3`, so the client bump drags a proto bump with it. Bumped both. It also removed `TransactionRequest::declares_fee_conversion_info`, which two of our tests asserted — ported to `fee_conversion_salt().is_none()`, which is the same property under the salt API. Both would have been silent until CI.

Stack is green: 1243 Rust, 626 TS, clippy `-D warnings`, fmt, browser determinism gate.

crates/client/Cargo.toml
```diff
@@ -11,7 +11,7 @@ description = "Client library for Guardian"
 include = ["src/**/*", "proto/**/*", "build.rs", "Cargo.toml", "LICENSE", "README.md"]
 
 [dependencies]
-guardian-shared = { version = "=0.17.0-rc.2", path = "../shared" }
+guardian-shared = { version = "=0.17.0-rc.3", path = "../shared" }
```
zeljkoX, line 14: We are bumping versions as part of release commit. No changes needed, given we will release once those PRs are merged.

zeljkoX (approved):
Thanks for working on this. Really good optimization on ts side supported by new version of web-sdk.

Only nits that are not blocking and will be resolved separately via its own tickets:
- ts formatting
- release version bump

We can merge this one.

PR #451, opened 2026-09-03
Pay the guarded-account fee from the Rust client

WiktorStarczewski:
The Rust client declares a fee-conversion salt on every guarded-multisig request, so a guarded account can transact on a chain whose `verification_base_fee` is non-zero.

- The five typed builders call `.fee_conversion_salt(salt)`; miden-client derives the conversion info from the anchored block's fee faucet and commits `hash(CONVERSION_INFO || SALT)` into the auth arg itself.
- `crates/contracts/tests/auth/fee_payment.rs` adds four kernel-level tests: the fee is paid, paid alongside a user note, a bare auth arg is rejected at non-zero fee, and a reversed preimage is rejected.
- `procedures.rs` asserts the built account classifies as `AuthGuardedMultisig` — see the note below on why that test is the load-bearing one.

<details>
<summary>Why the client commits it, and not this crate</summary>

rc.4 classifies an account's auth component and acts on it: `AuthGuardedMultisig` maps to `FeeAuth::CallerChosenSalt`, which requires the caller to declare a salt and then commits the native 1/1 conversion info against the **anchor's** fee faucet. `prepare_transaction_inner` takes `reference_header` from the supplied anchor, and every rebuild path here supplies the proposal's `chain_anchor`, so a cosigned summary reproduces exactly.

An earlier revision of this PR built the commitment by hand — an extension trait over `TransactionRequestBuilder`, an `Option<FeeConversionInfo>` threaded through every builder, and faucet resolution in `execution.rs`. That was a workaround for an account the client could not classify, which is a problem #450 removes: accounts now come from the upstream `AuthGuardedMultisig` component, so the client recognises them and does the work. The hand-rolled path is deleted (net −1100 lines).

**This makes #450 a blocking dependency rather than a convenience.** Under the old shape a misclassified account still got a correct commitment. Now, `FeeAuth::Ignored` plus a declared salt is a hard `FeeConversionInfoUnsupported` — every guarded transaction on a fee-charging chain fails, not just the fee. `rust_built_accounts_classify_as_guarded_multisig` is what stands between a standards bump and that failure: miden-client matches on the component's **full** procedure-root set, so pinning `auth_tx` alone would stay green while classification silently broke.

</details>

<details>
<summary>What the kernel tests do and do not prove</summary>

`fee_payment.rs` runs on `MockChainBuilder` + `miden_tx::TransactionExecutor` and never touches miden-client. It proves the MASM accepts the commitment shape, pays the fee, and rejects both a reversed preimage and a bare salt. It is deliberately **not** evidence about which side of the SDK boundary built the auth arg — the client's `commit_native_fee_conversion_info` emits byte-identical artifacts — which is why the classification test carries that weight instead.

</details>

**One local-only failure:** `cargo test -p miden-multisig-client --lib` fails on `client::note_recovery::tests::recover_notes_continues_past_failing_steps`. That test asserts the direct node channel is unreachable, so it fails on any machine running a local `miden-node` — its `scanned_to` tracks the live chain tip. Green in CI on `main` and untouched by this work. Otherwise green: 217 client tests, the four contracts fee tests, clippy `-D warnings`, `cargo fmt`.

Second of four stacked PRs. Based on the upstream-component PR — the rc.9 pin and the 1.98 toolchain live there, and nothing here builds without them.

**Reviewers:** `procedures.rs::rust_built_accounts_classify_as_guarded_multisig`. The whole design now rests on that classification holding.

---

<sub>

**The stack** — each PR targets its parent, so the diff shown is that step alone.

| | | |
|---|---|---|
| 1 | Build guarded accounts from the standard component | [#450](https://github.com/OpenZeppelin/guardian/pull/450) |
| 2 | Pay the guarded-account fee from the Rust client | [#451](https://github.com/OpenZeppelin/guardian/pull/451) |
| 3 | Commit fee conversion info in the TypeScript request builders | [#452](https://github.com/OpenZeppelin/guardian/pull/452) |
| 4 | Commit the anchored block's fee faucet on every proposal | [#453](https://github.com/OpenZeppelin/guardian/pull/453) |

Retargeted to `main` as each parent merges. [#440](https://github.com/OpenZeppelin/guardian/pull/440) carries all four as one diff with the original review threads.

</sub>


haseebrabbani:
- **The custom-path rationale is wrong on rc.9.** `helpers.rs:524-537` says a stale faucet "collects signatures and fails to pay" and that a non-native faucet is a legitimate case a seal would reject. The guarded auth calls `fee::assert_fee_bound`, which asserts the payment faucet equals `tx::get_fee_faucet_id`, so a stale read aborts inside `propose_custom_transaction`'s own `execute_for_summary` and a non-native faucet is never accepted. The seal is possible (compare the advice preimage's faucet to the captured anchor's). Add it or fix the comment.
- **Breaking for pending proposals, undocumented.** The rebuild now commits unconditionally, so any typed proposal pending from before the upgrade fails `verify_proposal_summary_binding` with "metadata does not match tx_summary", and `list_proposals` errors for the whole list. Fold a "drain pending proposals" step into the breaking-release note #450 needs.
- **Docs.** `docs/MULTISIG_SDK.md` Producer API still describes the custom recipe as inputs plus salt. `examples/demo/README.md` lacks the fee-funding prerequisite `examples/rust/README.md` gained. `payment.rs:33` and the crate README say a bare salt "aborts at proving"; it aborts at execution.

WiktorStarczewski:
All three fixed.

**Custom-path rationale.** You're right, and I verified it: `fee/mod.masm:279` does `exec.tx::get_fee_faucet_id` then `assert.err=ERR_FEE_PAYMENT_FAUCET_NOT_NATIVE`, and `guarded_multisig.masm:113` calls it. So a non-native faucet is never accepted — which removes the "legitimate case a seal would reject" the comment leaned on — and a stale read aborts inside this method's own `execute_for_summary`, before cosigners see it. Comment rewritten to say that, and to note the seal is expressible (compare the advice preimage's faucet to the captured anchor) rather than precluded. Not adding it here; that's a behaviour change and this PR is already the fee-payment one.

**Pending proposals.** Folded into #450's breaking-release section as its own "Drain pending proposals before upgrading" note, including the part your comment made explicit and mine didn't: the blast radius is the whole account, since strict `list_proposals` / `syncProposals` fail while GUARDIAN keeps serving the proposal, and the isolating recovery sync skips it silently so its notes never import.

**Docs.** Producer API now states the custom path must commit the info itself, with both error codes. `examples/demo/README.md` gained the fee-funding prerequisite. "Aborts at proving" corrected to execution in `payment.rs` and the crate README — and across the other six sites in #453, since it was wrong everywhere.

crates/contracts/tests/auth/fee_payment.rs
```diff
@@ -0,0 +1,404 @@
...
+use miden_tx::TransactionExecutorError;
+use miden_tx::auth::{BasicAuthenticator, SigningInputs, TransactionAuthenticator};
+use rand::SeedableRng;
+use rand_chacha::ChaCha20Rng;
+
+const NUM_APPROVERS: usize = 2;
+const VERIFICATION_BASE_FEE: u32 = 500;
+const FEE_FUNDING_AMOUNT: u64 = 1_000_000;
+/// Which linkage of the shared auth MASM the account under test carries.
+#[derive(Debug, Clone, Copy)]
```
zeljkoX: Due to the changes in #450, I don’t think this distinction is valid or necessary anymore, since both paths now use the upstream-compiled AuthGuardedMultisig component.

zeljkoX (approved):
LGTM

Thanks!

PR #452, opened 2026-09-03
Commit fee conversion info in the TypeScript request builders

WiktorStarczewski:
The TypeScript request builders declare the proposal salt through `withFeeConversionSalt`, so a guarded account can transact on a chain whose `verification_base_fee` is non-zero. Mirrors #451 on the Rust side.

- The five typed builders call `withFeeConversionSalt(salt)`; miden-client derives the native 1/1 conversion info from the anchored block's fee faucet and commits `hash(CONVERSION_INFO || SALT)` into the auth arg itself.
- The proposal salt is now **required** wherever a request is rebuilt — see below, this is the part worth reading.
- `tests/fee-conversion-salt-wiring.test.ts` checks the real WASM stores the salt, leaves the auth arg unset, and carries it across serialization.

<details>
<summary>Why the salt became mandatory</summary>

`summaryAuthArg(summary)` used to double as salt recovery, and it was correct: with no fee faucet supplied the auth arg **was** the bare salt. Once the request declares a salt, miden-client always commits — a declared salt bypasses its `verification_base_fee == 0` early return — so the auth arg is the commitment on every chain, and a commitment is not invertible.

That fallback is reachable: `salt` is optional on the wire, GUARDIAN stores the payload opaquely, and `fromGuardian` never checked it. For the six reconstruct-and-compare types it failed closed with a misleading "metadata does not match tx_summary". For `switch_guardian` it failed **open** — the binding check returns early above the salt line, so a wrong salt reached `buildUpdateGuardianTransactionRequest` and surfaced as an opaque MASM auth abort at execute time, after the pre-switch note drain had already run.

`prepareProposalExecution` now throws a named error instead. `ProposalMetadata::salt()` on the Rust side did the same thing more quietly — it answered `Word::default()` — and now errors too.

Deliberately **not** enforced in `ProposalMetadataCodec.validate()`: that runs for every proposal the factory builds, including ones only being listed or exported, and `exportProposal()` is the one path that still works for a salt-less proposal because it reads GUARDIAN's copy without rebuilding. Rejecting at the boundary would remove the only recovery route.

</details>

<details>
<summary>What the swap replaced, and the handle it leaked</summary>

`src/transaction/feeAuth.ts` (205 lines) computed the commitment, set `withAuthArg`, and extended the advice map. That was the escape hatch the SDK documents for "an account whose auth component miden-client does not recognise" — which is what #450 fixes, so the hatch is no longer needed.

`withFeeConversionSalt` borrows its argument exactly as `withAuthArg` did (`_assertClass` then `__wbg_ptr`, no `__destroy_into_raw`). The deleted `applyAuthArg` freed the salt in a `finally`; the five call sites now do it themselves, or every built request leaked a `Word`.

`SignatureOptions.feeFaucetId` is gone. It let a caller commit under an arbitrary faucet, which its own doc already described as unrecoverable — the client always uses the chain-native asset at 1/1. No published release exposed it.

</details>

**Rebased onto #451.** That is a precondition, not tidiness: this PR's docs claim both SDKs declare the salt, which is only true once #451's Rust swap is in the base.

**One local-only failure:** `cargo test -p miden-multisig-client --lib` fails on `client::note_recovery::tests::recover_notes_continues_past_failing_steps`, which asserts the direct node channel is unreachable and so fails on any machine running a local `miden-node`. Otherwise green: 589 TS tests, 217 Rust, `tsc`, clippy `-D warnings`, `cargo fmt`.

Third of four stacked PRs.

**Reviewers:** the salt-fallback removal. `switch_guardian` failing open is the case that motivated it.


haseebrabbani (approved):
minor comments:

- **Three exported error classes that nothing throws.** `ProposalAuthArgUnresolvableError`, `ProposalSaltMalformedError`, `FeeFaucetAnchorMismatchError` are on the package root and pinned by `public-api.test.ts`, but no `src/` path constructs them until #453. Move `authArgErrors.ts` there or state that the surface is intentionally ahead of its raiser.
- **Inconsistent sanitisation in `authArgErrors.ts`.** `ProposalSaltMalformedError` bounds and strips the untrusted salt via `quoteSalt`, but its `proposalId`/`reason` and every GUARDIAN-served field of `ProposalAuthArgUnresolvableError` are interpolated raw. Quote all of them or drop the claim.
- **Resolve CodeRabbit's `multisig.ts:1948` thread with a pointer.** It is valid: `prepareProposalExecution` and the `:2262` fallback still use `summaryAuthArg` as the salt, and the README still imports the removed `summarySalt`.

WiktorStarczewski:
All three addressed.

**Error surface ahead of its raiser.** Confirmed — nothing in `src/` constructs them here, all three arrive in #453. Kept the file where it is and documented the intent in a module header: landing the codes first keeps them out of #453's diff and lets a caller pin `code` before the raiser exists, with `public-api.test.ts` covering the barrel either way. Moving it would just shift the same lines into a larger PR.

**Sanitisation.** Fixed properly rather than narrowly: `quoteSalt` is now `quoteUntrusted` and applied to every GUARDIAN-served field these errors interpolate — `proposalId` and `reason` on `ProposalSaltMalformedError`, all four on `ProposalAuthArgUnresolvableError`, and both faucets on `FeeFaucetAnchorMismatchError`, which had the same hole. Stored properties stay raw so callers still get the real values; only the message is quoted.

**CodeRabbit `multisig.ts:1948`.** Valid, and the code half is already handled in #453 — the misleading `const saltHex = summaryAuthArg(...)` is gone and the fallback became the named `summaryBareAuthArg`, documented as a path that verifies but may not execute. The README was the live half: it still told integrators to recover the salt via `summarySalt`. That is worse than a rename — once the auth arg is `hash(CONVERSION_INFO || SALT)` it is not invertible at all — so the recipe now says to retain the salt or read `saltHex` from metadata, and shows `summaryAuthArg` for what the summary does still yield.

zeljkoX:
Just to double check, does `withFeeConversionSalt`  and proposed changes with https://github.com/OpenZeppelin/guardian/commit/f1092fac72ace3c31a87c6d8acf99b54980d657d make sense?

WiktorStarczewski:
rebased onto #451 (needed — the "both SDKs declare the salt" doc line is only true once the rust swap is in the base), then folded in @zeljkoX's f1092fac as-is.

zeljkoX (approved):
LGTM

Thanks!

PR #453, opened 2026-09-03
Pin the re-fetched summary, validate the served salt, release its handle

WiktorStarczewski:
Three defects on the proposal-rebuild path, none of them about fees. This PR used to thread the anchored block's fee faucet through every creator; miden-client does that itself since #451/#452, so that work is deleted and what is left is the hardening that was riding along inside it.

- **The re-fetched `tx_summary` is now pinned to the proposal id.** `executeProposal` re-fetches the summary from GUARDIAN rather than using the one already bound to the id, then keys the advice map and rebuilds the request from it. `ensureProposalCommitmentMatchesSummary` only ever pinned the *cached* summary, so nothing checked this one.
- **A served salt is validated, not just tested for truthiness.** `normalizeHexWord` left-pads, so `'0x'` and `'0X'` were truthy and padded to the **zero word** — a salt nobody chose. Non-strings and over-length values are rejected too, and `getP2idNoteId` goes through the same check.
- **The salt's WASM handle is released.** The builders read `.toHex()` and allocate their own `Word`, so the caller's handle was orphaned once per proposal per sync and once per execute.

<details>
<summary>What this PR used to be, and why almost none of it survived</summary>

The original version added `getFeeFaucetId()`, threaded a `feeFaucetId` through all seven creators and the rebuild paths, and sealed the committed faucet into the anchor. Every one of those symbols has since been deleted from this PR's own base: miden-client derives the faucet from the execution reference header, which is the proposal's anchor, so there is nothing left to thread.

`sealProposalAnchor` is the clearest case. It asserted that the anchor's faucet matched the committed one — but `executeForSummary` captures one anchor, executes at it, and returns that same object to be serialized into metadata. The header the client derives `CONVERSION_INFO` from and the header shipped in `chainAnchor` are the same object, so the check cannot fail. The race it caught existed only because the old code read the faucet out of band.

The previous head is `b4667dc` if any of it is wanted back; #440 also carries the original diff with its review threads.

</details>

<details>
<summary>Two gaps this does not close</summary>

**`switch_guardian` has no salt binding.** `verifyProposalMetadataBinding` returns early for that type before any reconstruction, so the *outgoing* GUARDIAN serves the salt for the proposal that removes it and nothing checks the value. The validation above narrows the shapes it can pick, but detect-and-refuse — recompute `hash(CONVERSION_INFO || SALT)` from the served salt and the anchor's faucet and compare against the summary's auth arg — is a separate change. The old fallback that "tried anyway" is not portable: `withFeeConversionSalt` owns the auth arg and `withAuthArg` is mutually exclusive with it, so there is no way to install a bare word any more. That is a real capability regression and worth an explicit decision rather than silent loss.

**Nothing pins "the anchor, not the tip".** The code is right at all four sites, but the anchor is now the *sole* determinant of the committed fee faucet, and `multisig.test.ts` never asserts that `submitAt` / `executeForSummaryAt` receive the anchor decoded from the proposal.

</details>

Green: 593 TS tests, `tsc`. Each new guard was sabotage-checked — removing it fails exactly the test that names it.

**Not claimed:** this stack has not been re-verified end to end. The earlier localnet run cited on the original PR does not transfer: every request then carried an auth arg, and miden-client's `attach_native_fee_conversion_info` returns early when one is set, so the classification path that now runs on every guarded transaction was never exercised by it.

Last of four stacked PRs.

**Reviewers:** the re-fetched-summary guard. On a Falcon roster nothing else compares that summary to the id.


haseebrabbani (approved):
minor comments:

- **Salt handle leaks on the two new paths.** `prepareProposalExecution` allocates `executionAuthArg.salt` at ~2141 but only consumes it at ~2249; every throw in between (signature checks, `pushDelta`, missing GUARDIAN ack, endpoint commitment) leaks it. In `recoverProposalAuthArg` the parsed salt is freed only on the mismatch branch, so a throw from `proposalFeeFaucetIdHex`, `nativeConversionInfo` or `feeAuthArg` leaks it too. Wrap both in `try/finally`.
- **Commit `f9393ec6` is not what it says.** "drop three feeAuth assertions that cannot fail" only changes quote style in `feeAuth.test.ts` (12 `expect()` before and after), leaving it the lone double-quoted test in the package. Revert to the base blob, or actually drop the assertions.
- **Pre-upgrade pending proposals now fail hard; put it in the release note.** `verifyProposalMetadataBinding` requires `hash(CONVERSION_INFO || salt) == signed auth arg`, so any proposal from 0.17.0-rc.1/rc.2 throws `proposal_auth_arg_unresolvable`, strict `syncProposals` fails the whole account, and the isolating recovery sync silently skips them so their notes are never imported. Matches Rust after #451, but "drain pending proposals before upgrading" has to land with #450's breaking-release row.
- **"Aborts at proving" is wrong where operators will grep for it.** `TROUBLESHOOTING.md` opens with "execution aborts during proving"; the README and `summaryBareAuthArg` doc say the same. The guarded auth runs during execution, so it surfaces from `executeForSummary`/`submitTransaction` as an execution error. `MULTISIG_SDK.md` in this same PR gets it right.

WiktorStarczewski:
Three fixed, one already resolved.

**Salt handle leaks.** Real, both of them. `recoverProposalAuthArg` freed the salt only on the mismatch branch, so a throw from `proposalFeeFaucetIdHex`, `nativeConversionInfo` or `feeAuthArg` leaked it — now wrapped so every exit that isn't the hand-off frees it, and the mismatch branch no longer needs its own free. `prepareProposalExecution` leaked across everything between the auth-arg build and `buildTransactionRequestFromMetadata` — signature checks, the GUARDIAN round trip, the endpoint commitment — now wrapped the same way.

**Commit `f9393ec6`.** You caught this on a commit that was briefly pushed and is now superseded. You were right about it: it was pure quote churn from a stray `prettier --write` (this repo has no prettier config, so it applied its own defaults over the house style). The pushed commit is a clean `30 deletions, 0 insertions`, removing 3 `expect()` and adding none — the three assertions really are gone.

**Pre-upgrade pending proposals.** Landed in #450's breaking-release section as its own note, with the whole-account blast radius and the silent-skip-in-recovery-sync detail spelled out.

**"Aborts at proving."** Wrong in eight places, not three — `TROUBLESHOOTING.md`, both READMEs, `options.ts`, `payment.rs` and four sites in `multisig.ts`. All corrected to execution, and `TROUBLESHOOTING.md` now says why once, where operators grep: the guarded auth runs during execution, so it surfaces from `executeForSummary` / `submitTransaction` before a proof is attempted.

Stack green throughout: 1246 Rust, 637 TS, clippy `-D warnings`, fmt, browser determinism gate.

WiktorStarczewski:
rewrote this one rather than rebasing it. after #451/#452 miden-client derives the anchored block's fee faucet itself, so `getFeeFaucetId`, the `feeFaucetId` threading and `sealProposalAnchor` had nothing left to do — all of it deleted (~2000 lines). previous head `b4667dc` if any of it is wanted back.

what was left underneath turned out to be three real defects unrelated to fees, so the PR is now those:
- `executeProposal` re-fetches the tx_summary from GUARDIAN and never checked it against the proposal id (the existing check pins the *cached* one). on a falcon roster nothing else compares them.
- `'0x'` / `'0X'` were accepted as salts and padded to the zero word.
- the salt's WASM handle leaked once per proposal per sync.

two gaps left open and written into the body rather than quietly dropped: switch_guardian still has no salt binding, and nothing pins "anchor, not tip" now that the anchor is the sole determinant of the committed faucet.

zeljkoX (approved):
LGTM

PR #455, opened 2026-09-04
451 simplification

zeljkoX:
#451 simplification

## Summary by CodeRabbit

- **New Features**
  - Multisig transaction builders now handle fee-conversion salts correctly across payments, note consumption, guardian changes, and configuration updates.
  - Added fee-payment coverage for guarded multisig accounts on fee-charging networks.
  - Account creation now uses the upstream guarded multisig implementation for consistent Rust and TypeScript results.

- **Bug Fixes**
  - Improved proposal execution by using the correct chain anchor during transaction construction.
  - Clarified handling of fee-related authentication data and transaction reconstruction.

- **Documentation**
  - Added guidance on funding accounts before transactions on fee-charging networks.
  - Documented compatibility with Miden 0.16 and updated SDK setup instructions.


WiktorStarczewski:
folded into #451 (2eb6b5a) — cherry-picked your commit as-is, kept your authorship. thanks, the salt path is much better.

one thing i added back: the classification test. dropping `assert_commits_fee_faucet` + the client fixtures leaves nothing checking that the account actually classifies as `AuthGuardedMultisig` — and with salt-only requests that's now load-bearing, since `FeeAuth::Ignored` + declared salt is a hard `FeeConversionInfoUnsupported`, not a missing commitment. the builder tests only prove a setter got called, and `rust_built_accounts_carry_the_pinned_auth_tx_root` only covers one root while `extract_component` needs all seven.

so `procedures.rs` now asserts `from_procedures` returns `AuthGuardedMultisig`. sabotaged it to confirm it discriminates.

also: my #451 body was wrong about why we hand-rolled it — claimed rc.4 rejects `AuthGuardedMultisig` outright. it doesn't, it's in the `CallerChosenSalt` arm. real blocker was the unclassifiable account, which #450 fixes. body corrected.

ok to close this one?

WiktorStarczewski:
same treatment for #452 — folded f1092fac into split/pr3-ts-fee-auth (fb7df6f), your commit kept as-is. `withFeeConversionSalt` confirmed on rc.7, and the TS account is bit-identical to the rust one (same id/commitment/storage as the parity test), so classification holds on both sides.

two things i added on top:

1. **the salt fallback had to go.** `summaryAuthArg(summary)` as salt recovery was correct before — with no feeFaucetId the auth arg WAS the bare salt. once we declare a salt the client always commits (a declared salt skips the `verification_base_fee == 0` early return), so it's the commitment on every chain and not invertible. reachable too: `salt` is optional on the wire and `fromGuardian` never checked it. six types fail closed with a misleading "metadata does not match tx_summary"; **switch_guardian fails open** — binding check returns early above the salt line, wrong salt reaches the builder, opaque MASM abort at execute, after the note drain. now throws by name. did the same to rust `ProposalMetadata::salt()`, which was quietly returning `Word::default()`.

2. **`withFeeConversionSalt` borrows, same as `withAuthArg`** — `_assertClass` then `__wbg_ptr`, no `__destroy_into_raw`. `applyAuthArg` used to free the salt in a finally; the five new call sites didn't, so every built request leaked a Word. freeing at each site now.

also put back a real-WASM test (`tests/fee-conversion-salt-wiring.test.ts`) — after the swap the only fee coverage left was mocks we wrote ourselves, and the deleted `tests/fee-auth-request-wiring.test.ts` existed precisely because a fake builder can't tell you whether the SDK stored anything.

one i did NOT take your way: not enforcing salt in `ProposalMetadataCodec.validate()`. it runs for every proposal incl. list/export, and `exportProposal()` is the only thing that still works for a salt-less one. rejecting there kills the last recovery path.

note i rebased pr3 onto pr2 — your doc line "both SDKs declare the stored proposal salt" is false until the rust swap is in the base.

ok to close #455?

PR #458, opened 2026-09-07
fix(multisig-client): reconcile the proposal cache with GUARDIAN on sync

### LINKED ISSUE #404: syncProposals() keeps returning executed proposals as pending (stale cache never pruned)

zeljkoX:
### What

`syncProposals()` reconciles the local proposal cache against the GUARDIAN response on every sync instead of only ever adding to a Map that was never cleaned up.

This supersedes #405 by @WiktorStarczewski, which went stale against main (conflicting) with two review findings unaddressed. The core design (prune relative to what GUARDIAN reported, no client-side nonce filter, TS-only scope since the Rust listing is fetch-fresh) is Wiktor's; credit for the diagnosis and the original fix goes to him.

### Why

The cache was merge-only. After a proposal is executed, GUARDIAN stops returning it, but every non-executing signer's client silently ignored the reduced response and kept the executed proposal pinned as pending forever, stuck at e.g. 1/2 (#404).

### Pruning contract

Pruning is driven by whether GUARDIAN provably holds a proposal (`guardianKnownProposalIds`: acknowledged `createProposal` pushes, acknowledged online `signProposal` signatures, and every id a listing has reported):

- A proposal GUARDIAN reported and then stopped reporting is pruned on the first listing that omits it.
- A guardian-known proposal not yet listed (a fresh create during read-your-writes lag, or one orphaned by a `setGuardianClient` repoint) is pruned only after two consecutive listings omit it, so the #404 creator path converges by the second sync.
- Proposals GUARDIAN never received (`importProposal`, `createSwitchGuardianProposalOffline`) are never pruned by listings; an import graduates into the pruned classes once GUARDIAN acknowledges it (listed, or signed online).

### Concurrency and failure semantics

- The whole response is verified into a temporary structure before the cache or pruning state changes; a listing that fails metadata-binding verification throws and leaves both untouched.
- Overlapping callers share the single in-flight promise (by identity; `syncProposals` is not `async` for this reason).
- The apply phase rebuilds each reported proposal from the current cache signatures, so a signature added while the sync was verifying survives.
- Cache entries created or replaced after the sync started are not evaluated by that sync (entry-identity snapshot), so a proposal re-acknowledged mid-flight cannot be deleted by a stale listing.
- A sync that spans a `setGuardianClient` repoint rejects without applying its listing, and the repoint resets the reported/miss bookkeeping.

Nonce-based staleness hiding stays with the callers' `filterVisibleProposals`: consumers disagree on the proposal `nonce` convention, so the Rust-style `nonce <= account nonce` filter cannot live in the shared client. Unifying that convention is the follow-up that would retire most of this bookkeeping.

### Tests

Sync-focused tests cover: co-signer prune, over-prune guard, creator-path convergence, offline import retention, offline switch-guardian retention, import graduation via online signing, verify-failure atomicity (empty and seeded cache), mid-verify signature preservation, stale-listing versus mid-flight sign, overlap dedupe with promise identity, repoint abandonment, and no cross-guardian pruning. Each guard is mutation-checked to fail when its mechanism is removed. Full package suite green, `tsc --noEmit` clean.

### Docs

README sync section, `docs/MULTISIG_SDK.md` method rows, and the method JSDoc state the reconcile semantics, including the intentional TS/Rust surface difference.

### Release / semver note

This changes the observable output of `syncProposals()` / `listProposals()`: they no longer retain proposals GUARDIAN has dropped, and a pushed-but-never-listed create is pruned after two consecutive omissions. It fixes buggy behavior and the in-repo example consumers re-derive their visible set each sync, but the release carrying it should flag this in the notes. No version bump included; versioning is a separate release commit.

Closes #404


haseebrabbani (approved):
LGTM

Issue #462, opened 2026-09-09
Pending proposals become unverifiable and unexecutable once the node prunes the anchor block (fee-faucet foreign load at the reference block)


haseebrabbani:
### Summary

Every guarded-multisig transaction loads the fee faucet as a foreign account, and anchored re-execution fetches that foreign account at the proposal's reference block. Public nodes serve historical account state only for a short window, so once the anchor block is pruned nobody, including the proposer, can verify or execute the pending proposal, and the strict listing then aborts on it for every other proposal of the account.

Found while reproducing #409 live on staging (0.17.0-rc.3, devnet). This is independent of the note-authentication fix for #409 and is not addressed by it.

### Reproduction (staging + devnet, Rust demo)

1. Create a 2-of-2 account, fund it from the devnet faucet, create any proposal (a consume-notes proposal was used; the mechanism applies to every type).
2. Wait a few minutes.
3. List proposals from any client, including the proposer:

```
✗ Error: Failed to fetch proposals: transaction execution failed: failed to execute transaction kernel program:
|-> error during processing of event 'miden::protocol::account::before_foreign_load'
|-> failed to get inputs for foreign account 0x157e8ac22390f771044593acdc153f from data store at reference block 4502
|-> failed to fetch foreign account inputs
|-> RPC error
|-> grpc request failed for get_account: invalid request parameters
`-> code: 'Client specified an invalid argument', message: "block 4502 has been pruned"
```

`0x157e8ac22390f771044593acdc153f` is the devnet fee faucet.

### Measured retention

A probe calling `get_account` for the fee faucet at `AccountStateAt::Block(tip - k)` against `rpc.devnet.miden.io` (2026-09-08): `k = 50` served, `k = 52` and beyond refused with "has been pruned". At roughly 3 s per block that is about 2.5 minutes from proposal creation to the point where the proposal is dead. Testnet retention not measured.

### Mechanism

- The kernel's asset callbacks (`callbacks.masm`, `on_before_asset_added_to_account` / `on_before_asset_added_to_note` → `maybe_start_faucet_callback_context`) check whether the asset's faucet defines a callback. For that they start a foreign context against the faucet, which emits `before_foreign_load` and loads the faucet as a foreign account. Paying the fee moves the native asset, so the fee faucet is loaded on every fee-paying transaction. This is protocol-level; nothing in Guardian's contracts or request builders asks for it.
[...]

PR #463, opened 2026-09-09
fix(multisig-client): list unverifiable proposals instead of failing the whole listing (#462)

haseebrabbani:
### What

Once the node prunes a proposal's anchor block, re-executing that proposal fails for everyone: the kernel's asset callbacks load the fee faucet as a foreign account at the reference block, and devnet serves only ~50 blocks (~2.5 min) of account history. The strict listing aborted on the first such proposal, so one stale proposal hid every other proposal and the account could not be loaded at all (#462).

### Change

- `Proposal` gains `verification_error` (Rust, with `is_verified()`) / `verificationError` (TS). The strict listing (`list_proposals`, `syncProposals`) returns every parsed proposal and records a failed summary-binding check on the proposal instead of failing. A payload that does not parse still fails the listing, so malformed GUARDIAN data is never silently dropped.
- Signing and executing are unchanged: they re-verify the one proposal they act on and refuse it with the real error. The relaxation is listing-only.
- The demo marks unverifiable proposals in every listing. Docs show the field in both SDK examples and add an "Anchor lifetime" note on the pruning window and re-proposing.

### Not a fix for pruning

A pruned proposal stays unusable until proposals can carry the foreign-account inputs themselves (0xMiden/rust-sdk#2524). Until then: collect signatures and execute promptly, and re-propose once a proposal has aged out.

Refs #462.

## Summary by CodeRabbit

- **New Features**
  - Proposal listings now retain unverifiable proposals and show a `verificationError` explaining the issue.
  - Valid proposals remain available alongside unverifiable ones.
  - Verification uses each proposal’s chain anchor, improving listing across different sync heights.
  - Added checks to identify whether a proposal is verified.

- **Bug Fixes**
  - Proposal synchronization no longer fails entirely when individual proposal verification fails.
  - Unverifiable proposals remain blocked from signing and execution.

- **Documentation**
  - Added guidance on chain-anchor expiration and handling unverifiable proposals.


zeljkoX (commented):
I like where this lands.  Two things I'd want fixed before merge, and one design point that I think is worth a follow-up commit here rather than a later PR.

The #409 tests don't test anything anymore. Before this change a failed re-execution at the anchor threw out of the listing. Now it just lands in `verification_error` and the proposal comes back like the others, so both tests pass whether the anchored re-execution works or not. The Rust one (`anchor_binding_tests.rs:137-146`) needs an `assert!(proposals[0].is_verified())`, and the TS "#409" case needs `expect(proposals[0].verificationError).toBeUndefined()`. Cheap, but without it the regression guard is gone.

The demo still prints `✓ READY` from `signatures_needed() == 0` (`proposal_management.rs:488`), so a fully signed pruned proposal shows READY one line above UNVERIFIABLE. Gate the marker on verification too.

On the design. Keeping verification out of `ProposalStatus` is the right call, since a proposal can be fully signed and dead at the same time and folding that into status would hide the signature count. What I don't like is `Option<String>` as the value. `None` currently means two different things. From the listing it means "I rebuilt the summary and it matched". From `Proposal::from`, `from_transaction`, and `to_proposal` it means "nobody checked". And a string can't tell "retry later" from "re-propose now", which matters because a node hiccup during the re-execution now returns a successful listing with every proposal flagged, and the demo tells the user none of them can be signed or executed. I'd go with

```rust
pub enum Verification {
    Verified,
    Failed { retryable: bool, message: String },
    Unchecked,
}
```

written by verify_proposal_summary_binding itself so Verified can only come from the check. retryable can come from the existing is_transient_rpc_error. On the TS side make it a required discriminated union, not verificationError?: string, otherwise we get the same ambiguity back plus drift from Rust. An is_actionable() (verified and ready) covers the convenience case and lets Ready keep meaning "threshold met".

Also let's document behavior and changes in multisig README files.

zeljkoX (approved):
LGTM

One item to address before merging:

**Re-verify the proposal returned after signing**

`sign_proposal()` returns a freshly parsed `Proposal::from(updated_raw)`, which initializes verification to `Unchecked`. Consequently, `is_actionable()` returns false even after the final signature, and the demo’s new post-signing READY message never appears. TypeScript re-verifies the signed response before returning it. Do the same here and add a regression asserting that the final-signature response is both verified and actionable.

PR #469, opened 2026-09-11
feat(dashboard): bounded `GET /dashboard/stats` for account and Miden vault aggregates (#371)

### LINKED ISSUE #371: Dashboard API: Add a bounded `/dashboard/stats` endpoint for account and Miden vault aggregates
## Problem

The cross-operator [guardian-dashboard](https://github.com/0xMiden/guardian-dashboard) reports, per Guardian server, account activity and composition plus assets under guard. Guardian does not expose these as a complete aggregate API, so the dashboard reconstructs them client-side:

- it pages the full account list (`GET /dashboard/accounts`) to compute accounts updated within 7/30 days, account composition, and paused/released counts (`app/api/accounts/stats/route.ts`);
- it calls `GET /dashboard/accounts/{id}/snapshot` once for every account whose `updated_at` falls within the last seven days, then aggregates the decoded vault (`app/api/accounts/asset-totals/route.ts`).

On a Guardian holding 2,379 accounts, this is a full-list walk plus approximately 1,086 snapshot requests per refresh.

The code-default **DEV** HTTP limits are 10 requests/second burst and 60 requests/minute sustained (`GUARDIAN_RATE_BURST_PER_SEC` and `GUARDIAN_RATE_PER_MIN`). Production Terraform currently sets 200/second and 5,000/minute. On a Guardian enforcing the 60/minute sustained budget, operators observed the walk being cut off after approximately 57 requests—the remaining requests in the window having been consumed by authentication and inventory reads—with `429` and `Retry-After: 60`.

The dashboard compensates with adaptive fetch ceilings and does not publish a new total while snapshot reads remain unattempted because of its ceiling or a `429`. This leaves operators seeing states such as:

> Assets (7d active): Calculating… 25 of 1,086

Under sufficiently tight limits, completion can take many minutes or may never converge across cold serverless instances. Non-rate-limit snapshot failures are currently treated as attempted and omitted, so the client also lacks authoritative coverage information for deciding whether a total is complete.

Raising rate limits is a per-operator mitigation. The underlying problem is the request volume: only Guardian can aggregate its stored states efficiently and report authoritative coverage.

A related failure already appears at this scale. Above the configured aggregate threshold, `/dashboard/info` declines to compute `accounts_by_auth_method` and includes it in `degraded_aggregates`. On a 2,379-account Guardian, the overview consequently reports that the server cannot compute the account breakdown precisely where server-side aggregation matters most.

## Current API surface

| Endpoint | Provides | Gap |
|---|---|---|
| `GET /dashboard/info` | `total_account_count`, `accounts_by_auth_method`, delta status counts, in-flight proposal count, and latest activity | No asset data or activity-window/lifecycle counts; auth-method aggregation degrades above a threshold |
| `GET /dashboard/accounts` | Paginated account summaries | Consumers must walk the inventory to compute cross-account counts |
| `GET /dashboard/accounts/{id}/snapshot` | One decoded Miden vault | Requires one request per account to aggregate assets |

## Proposal

Add `GET /dashboard/stats`, protected by the operator session and `dashboard:read` permission, returning account and asset aggregates in one request.

This issue specifies observable requirements rather than implementation. Guardian may maintain incremental aggregates, serve a periodically refreshed snapshot, or use another bounded mechanism. The existing background-refresher pattern in `crates/server/src/metrics/refresher.rs` may be useful, but is not required.

### Functional requirements

**FR-1 — Asset totals**

Return totals across eligible Miden account snapshots:

- fungible base-unit sums grouped by `faucet_id`;
- non-fungible asset counts grouped by `faucet_id`.

Fungible sums must be serialized as base-10 decimal strings and computed without `u64` or JavaScript safe-integer overflow.

Guardian must not perform decimals normalization, token pricing, or fiat valuation. Those remain consumer concerns.

**FR-2 — Activity filter**

Accept an optional RFC3339 `updated_since` query parameter for the asset aggregation.

For compatibility with the current dashboard, this filter refers to the account metadata `updated_at` exposed as `DashboardAccountSummary.updatedAt`, not the stored state row’s timestamp. Accounts are eligible when:

```text
account.updated_at >= updated_since
```

Omitting the parameter aggregates all accounts.

Invalid timestamps return the standard `400` error envelope with a stable error code.

**FR-3 — Account counts**

Return the following unfiltered account counts in the same response:

- total accounts;
- by mutually exclusive lifecycle state:
  - `released` when `released_at` is present;
  - otherwise `paused` when `paused_at` is present;
  - otherwise `active`;
- by stable auth-method label;
- by auth-method label and authorized-signer count, so the dashboard can reproduce its current account-shape heuristic without Guardian claiming that the shape identifies a particular client;
- accounts whose metadata `updated_at` falls within the last 7 and 30 days, anchored to the response’s `as_of`.

The endpoint must not silently omit auth-method counts above an inventory threshold.

**FR-4 — Coverage and degradation**

The response must include:

- `as_of`: the timestamp represented by the published aggregate;
- the applied `updated_since`, or `null`;
- the number of accounts eligible for asset aggregation;
- the number of accounts covered;
- the number of accounts skipped, grouped by a stable reason such as unavailable or undecodable state;
- an explicit `complete` indicator;
- stable names for any aggregate the server declined to compute.

Coverage must satisfy:

```text
covered + skipped = eligible
```

Any skipped eligible account makes the asset aggregate incomplete.

A degraded or unavailable aggregate must be represented explicitly. It must not be serialized as an empty map or zero that could be mistaken for a valid result.

**FR-5 — One-call answer**

`GET /dashboard/stats?updated_since=...` must answer the assets-under-guard and account-count questions without per-account follow-up requests.

**FR-6 — Bounded serving cost**

Serving the endpoint must not decode every eligible account vault on each request. Request-time work must remain bounded independently of inventory size at steady state.

Any background refresh must also be bounded or batched so that a large inventory cannot monopolize server resources. Staleness is acceptable when reported accurately through `as_of`.

**FR-7 — Freshness**

The refresh cadence or maximum expected staleness must be documented. The initial target should be no worse than the dashboard’s existing 60-second cache unless maintainers explicitly choose a different budget.

Consumers must be able to determine the age of the result from `as_of`.

**FR-8 — Authentication and errors**

- operator session required;
- `dashboard:read` permission required;
- `401` and `403` behavior identical to existing dashboard read endpoints;
- standard Guardian error envelope;
- ordinary HTTP rate limiting applies—one aggregate request replaces approximately 1,100 inventory/snapshot requests.

**FR-9 — Existing `/dashboard/info` degradation**

The implementation must prevent the existing `accounts_by_auth_method` threshold from continuing to report the server as degraded when the new stats aggregate is available.

This may be implemented by reusing the same maintained aggregate in `/dashboard/info`, or by moving that aggregate authoritatively to `/dashboard/stats` and updating consumers accordingly. The two endpoints must not report contradictory values.

**FR-10 — Contract propagation**

Update together:

- server handler, wire types, and tests;
- `#[utoipa::path]` and `ToSchema`/`IntoParams` declarations;
- generated `docs/openapi-dashboard.json`;
- `packages/guardian-operator-client`, including strict response parsing and tests;
- `examples/operator-smoke-web`;
- `docs/DASHBOARD.md`;
- relevant behavioral documentation in `spec/api.md`.

## Acceptance criteria

On a Guardian containing approximately 2,400 accounts, including approximately 1,000 whose metadata was updated within seven days:

1. `GET /dashboard/stats?updated_since=<7-days-ago>` returns asset totals, coverage, account totals, activity-window counts, lifecycle counts, and auth/account-shape counts.
2. A dashboard refresh requires at most two Guardian requests: `/dashboard/stats` and, if still needed, `/dashboard/info`.
3. No per-account snapshot requests are required for the overview.
4. The refresh succeeds under the code-default HTTP rate limits without `429` or a multi-minute calculating state.
5. Asset totals match a reference full-inventory snapshot aggregation for every covered account.
6. Missing or undecodable snapshots produce `complete: false` and explicit skipped coverage; they never appear as valid zero balances.
7. The operator TypeScript client and operator smoke harness exercise the new endpoint.
8. Filesystem and Postgres backends produce equivalent response semantics.

## Follow-ups

- **Per-account totals in list summaries:** the accounts table and CSV still need per-row totals. Adding fungible totals to `DashboardAccountSummary` or introducing a bulk endpoint should be tracked separately.
- **True client attribution:** the dashboard’s “wallet” classification is currently a heuristic based on auth scheme and signer count. A server-recorded client-attribution field at registration would be a separate contract change.

## Out of scope

- Changing the default rate limits.
- EVM accounts or asset aggregation.
- Token metadata, decimals normalization, asset pricing, or fiat valuation.
- Per-account asset totals for the accounts table and CSV.
- Introducing authoritative wallet/client attribution.

haseebrabbani:
Closes #371.

Adds `GET /dashboard/stats` (operator session + `dashboard:read`), answering the dashboard's "assets under guard" and account-composition questions in one request instead of a full account-list walk plus ~1,100 per-account snapshot reads.

**Server.** A per-replica background refresher (`dashboard/stats.rs`) walks metadata in pages of 200, batch-pulls Miden states, decodes a vault only when its commitment changed since the previous snapshot, and publishes an immutable snapshot every `GUARDIAN_DASHBOARD_STATS_REFRESH_INTERVAL_SECS` (default 60 s). Requests never touch storage. The response carries unfiltered account counts (total, lifecycle, auth method, auth-method × signer count, 7d/30d windows), Miden fungible totals as 128-bit decimal strings and non-fungible counts per faucet, and explicit coverage (`eligible`, `covered`, `skipped` by reason, `complete`). `?updated_since=<RFC3339>` filters the asset aggregate only; malformed values are `400 invalid_timestamp`. Until the first refresh the endpoint is `503 data_unavailable`; a failed refresh keeps the previous snapshot.

**FR-9.** `/dashboard/info` now serves `accounts_by_auth_method` and `total_account_count` from the same snapshot, so the inventory threshold no longer degrades that aggregate and the two endpoints cannot contradict each other.

**Propagation.** `#[utoipa::path]`/`ToSchema`/`IntoParams` + regenerated `docs/openapi*.json`; `getDashboardStats()` with strict parsing and tests in `guardian-operator-client`; new buttons in `examples/operator-smoke-web`; docs (`DASHBOARD.md`, `CONFIGURATION.md`, `spec/api.md`, `architecture/services.md`); metrics `guardian_dashboard_stats_refresh_{timestamp_seconds,failures_total,duration_seconds}`.

**Verified.** Full server suite (917), clippy/fmt/OpenAPI drift check, TS client tests (100), smoke-web typecheck, and a live run against the real binary with a mixed inventory (active/paused/released, missing state, corrupt state, EVM).

**Follow-ups (not in this PR).** Export the new metrics in `infra/observability.tf` + Grafana; add Postgres-backed coverage for the stats walk.

## Summary by CodeRabbit

* **New Features**
  * Added a dashboard statistics endpoint providing account counts, lifecycle and authentication breakdowns, signer counts, activity windows, and asset totals.
  * Added optional `updated_since` filtering for statistics.
  * Added client support and typed responses for retrieving dashboard statistics.
* **Improvements**
  * Dashboard information now uses consistent snapshot-based account aggregates.
  * Statistics refresh automatically in the background at a configurable interval.
* **Documentation**
  * Documented the endpoint, configuration, response fields, filtering, and error behavior.
* **Bug Fixes**
  * Added clear errors for unavailable data and invalid timestamps.


zeljkoX (commented):
I’m requesting an architecture change before merge. Issue #371 allows periodic snapshots, but for our multi-replica deployment I would prefer one shared published result. Currently, every replica repeats the aggregation independently, and requests routed to different replicas can return different totals and timestamps.

### Proposed architecture

1. **Persist the published aggregates in Postgres.** All replicas should read the same completed snapshot, identified by a version and `as_of`. Publish atomically. Preserve arbitrary `updated_since` filtering through persisted queryable data or cumulative aggregates; storing only the unfiltered response is insufficient.

2. **Reuse `worker_leases` and `PgLeaseElector` with a dedicated lease name.** Only the lease holder should refresh the data. Publication must prevent an older worker from overwriting a newer snapshot after leadership changes. A separate `verify_held` check before writing is insufficient; enforce fencing at publication and test the ownership-transfer race.

3. **Use a five-minute default background refresh interval.** This explicitly changes FR-7’s original freshness target and should be reflected in the issue and documentation. The interval must not expire the published data: continue serving the last successful snapshot during refreshes and failures.

4. **Allow operators to request a refresh.** Add an authenticated endpoint returning `202 Accepted`, with explicit permission requirements and a cooldown. Automatic and operator-triggered refreshes should share the same coordination mechanism. Requests received during an active refresh should return its status instead of starting duplicate work. Include OpenAPI, operator-client support, tests, and smoke-harness coverage.

5. **Have `/dashboard/info` reuse the shared inventory aggregates and expose their freshness.** I would include account/auth counts, delta status counts, in-flight proposals, and latest activity so the overview has a common snapshot time. Postgres currently computes the latter fields live, so this deliberately trades freshness for a consistent cached overview. Remove filesystem inventory-threshold degradation for fields successfully computed in the snapshot, while keeping genuine unavailability or failures explicit.

Filesystem deployments can retain a single-process in-memory implementation with equivalent coverage and failure semantics. Requests spanning a publication can still observe different versions; the shared snapshot eliminates differences caused by independent replica refresh schedules.

### Correctness and serving-cost fixes

These need addressing regardless of where the snapshot is stored:

- **Preserve the previous snapshot on storage-access failures.** The default `pull_states_batch` currently swallows non-missing read errors, allowing the refresher to publish reduced totals as a successful refresh. Propagate those errors and update the test that currently treats `"db down"` as a missing row.

- **Distinguish individual corrupt records from systemic failures.** One decryption failure currently aborts the entire refresh, leaving a newly started replica on `503` indefinitely until the problem is repaired. Individual corrupt records should produce explicit skipped coverage; systemic storage or encryption failures should retain the previous snapshot.

- **Retry failed decodes.** Caching `StateUndecodable` by commitment prevents recovery after a blob is repaired without changing its commitment. Cache successful decodes only, or periodically retry failures.

- **Bound filtered request work.** `updated_since` currently re-sums every matching account’s cached vault entries. Build the necessary query structure during refresh so the normal dashboard query does not require an inventory-sized fold.

- **Avoid repeated filesystem scans and sorts.** Filesystem `list_paged` scans and sorts the remaining inventory under the metadata read lock for every page. Use a single metadata snapshot or an indexed traversal.

Please add Postgres and encrypted-storage coverage for publication, failures, repair recovery, concurrent refresh requests, and lease loss. Also correct the freshness documentation: the configured interval is not a maximum snapshot age, particularly during slow or failed refreshes.

zeljkoX (approved):
Thanks

Looks good to me!

PR #476, opened 2026-09-15
feat: Allow overriding tf resources

zeljkoX:
## Summary

Lets operators keep Terraform state in a remote backend (S3 with lockfile, or any other backend that supports CLI workspaces; the `http` backend does not and is out of scope) without touching the tracked `infra/` module. Local per-stack state files stay the default; nothing changes for existing workflows unless an override file is added.

## Overlay contract

- The tracked module still declares no backend. An operator drops an untracked `infra/*_override.tf` (pattern already gitignored) carrying a `backend` block, and optionally an `assume_role` block on the AWS provider. Terraform merges override files into the configuration, so bucket names, account IDs and role ARNs never enter the repo. `infra/backend_override.tf.example` is the template.
- Two keys must be present together: a non-local backend block in an override, and `TF_WORKSPACE` naming the workspace for the stack. `scripts/aws-deploy.sh` refuses either half alone, because `TF_WORKSPACE` with a local backend would plan against fresh empty state, and a remote backend with `TF_WORKSPACE` unset would receive `-state=` flags that only apply to the local backend.
- `TF_WORKSPACE` only selects state. `STACK_NAME` and `DEPLOY_STAGE` still drive resource names and the deploy profile, so the docs pair all three in every snippet.

## Deploy script behaviour with both keys set

- Stops passing `-state=` and `-backup=`; Terraform resolves state through the backend and `TF_STATE_PATH` is ignored.
- Always runs `terraform init -reconfigure -input=false` so backend config changes are picked up and local state is never migrated implicitly.
- Selects the workspace and refuses if it does not exist, so a typo does not mint an empty workspace on every command.
- Refuses `deploy` and `cleanup` when the workspace has no resources in state, counting resource addresses from stdout only (stderr lock and log lines are not counted). `plan` only warns.
- Refuses to run when `infra/.terraform` was initialized with a non-local backend but no override declares a backend block anymore, instead of reconfiguring down to empty local state.

## `--bootstrap`

For a genuinely new stack only. Creates the workspace (`workspace select`, falling back to `workspace new`, so no Terraform 1.4+ flag is needed) and lifts the empty-state guard for that invocation. Not needed after the first successful deploy.

## Migration of existing local state

Documented in `docs/SERVER_AWS_DEPLOY.md` under "Remote state backend": init and create the workspace with `TF_WORKSPACE` unset, `terraform state push` the existing file with it set, then run `plan` with matching `STACK_NAME`, `DEPLOY_STAGE` and `TF_WORKSPACE` and confirm no changes.

## Caveats

- The script's own AWS CLI calls (ECR, Secrets Manager, STS) use ambient credentials, not the provider's `assume_role`. Cross-account roles need the shell authenticated to the same account.
- `use_lockfile` in the example raises the Terraform floor to 1.10 for operators who adopt it; the tracked module keeps `>= 1.0`.

## Testing

Exercised `ensure_terraform_init`, `remote_state_resource_count` and `require_remote_state_populated` against a stub `terraform` covering: populated, empty and missing state with stderr noise; workspace exists, typo without `--bootstrap`, new workspace with `--bootstrap`; override lost or reduced to provider-only while `.terraform` still points at s3; plain local path unchanged.

## Summary by CodeRabbit

- **New Features**
  - Added optional shared remote Terraform state support using `TF_WORKSPACE`.
  - Added workspace-aware deployment, status, plan, apply, and cleanup operations.
  - Added bootstrap support for creating missing remote workspaces and initializing empty state.
  - Added an example configuration for encrypted S3-backed state and locking.

- **Bug Fixes**
  - Added safeguards that prevent deployments or cleanup against empty or inconsistently configured remote state.

- **Documentation**
  - Expanded deployment and infrastructure guidance, including local-versus-remote state behavior, credential handling, migration guidance, and required environment variables.


haseebrabbani (approved):
Approved with two items to address:

1. **Return-to-local recipe cannot work.** Both `docs/SERVER_AWS_DEPLOY.md` and the "override lost" error in `ensure_terraform_init` tell the operator to run `terraform init -reconfigure`. With the backend block removed, Terraform 1.11 exits 0 but keeps `.terraform/terraform.tfstate` pointing at `s3`, so the script refuses again on the next run and there is no way out. Only `-migrate-state` unconfigures a backend, and it copies every workspace into `terraform.tfstate.d/`, which the script's `-state=` never reads. A recipe I verified end to end (plan reports no changes afterwards): `TF_WORKSPACE=<ws> terraform -chdir=infra state pull > infra/terraform.<stack>.<stage>.tfstate`, then remove the override and `rm -rf infra/.terraform`. Suggest updating both the docs and the error hint.
2. **Unquoted `$files` in `override_declares_backend`.** From a repo path containing a space, the first run passes (the check is skipped before backend metadata exists) and every later run refuses with "no override declares a backend block anymore" although the override is present. Quote the variable or loop over the files.

PR #448, opened 2026-09-02
ci: add OIDC-based AWS deploy workflow

haseebrabbani:
Closes #282.

Adds a `workflow_dispatch` workflow that deploys a published `ghcr.io/openzeppelin/guardian` release image to the stg or prod ECS stack, using GitHub OIDC role chaining only — no long-lived AWS keys.

### How it works
- Inputs: `environment` (stg/prod) and `version` (e.g. `v1.2.3`; prod refuses pre-releases).
- A `validate` job resolves the tag to an immutable digest and verifies that digest against the SLSA provenance attestation signed by `docker-publish.yml`; prod additionally requires the build to have run from `refs/tags/<version>`, so a manually dispatched build of another branch cannot ship to prod under a release version. This runs before environment protection rules, so prod reviewers only approve an already-verified digest.
- The `deploy` job (GitHub environment `stg`/`prod`) authenticates via OIDC, checks the ECR repo and ECS service exist, mirrors the verified digest into `<stack>-server` tagged both `<version>` and `latest`, registers a new revision of the task definition the service is currently running with only the image changed, and waits for stability.
- Moving ECR `latest` keeps `scripts/aws-deploy.sh plan` / `deploy --skip-build` resolving the deployed image, so a later Terraform apply does not roll the service back.
- The run summary always records the deployed image and previous task-definition ARN, and prints the rollback command on failure.

### Required setup (infra)
- GitHub environments `stg` and `prod` with variables `AWS_REGION`, `ROLE_FOR_OIDC`, `ROLE_TO_ASSUME`, `STACK_NAME` (`guardian` / `guardian-prod`); required reviewers on `prod` recommended.
- OIDC role trust policy extended to `repo:OpenZeppelin/guardian:environment:stg|prod`.
- Deploy role permissions are listed in `docs/SERVER_AWS_DEPLOY.md`.

### Notes
- Image-only rollouts: when a release also changes the task definition (env vars, secrets, IAM grants), apply that release's `infra/` first, then deploy the image.
- `guardian-evm` is out of scope: GHCR release images are built with `features=postgres` only.
- Follow-up worth filing: `deployment_circuit_breaker { enable = true, rollback = true }` on the ECS service so a bad image fails fast instead of timing out the 20-minute wait.
- Docs updated: `docs/SERVER_AWS_DEPLOY.md` (new section), `infra/README.md` (pointer).

PR #442, opened 2026-08-31
feat(infra): ADOT metrics sidecar with CloudWatch dashboard and alarms

haseebrabbani:
Closes #300.

Ships Guardian application metrics to CloudWatch in the AWS reference deployment, building on the Prometheus endpoint from #225.

## What's here

- **ECS**: the server task now enables the metrics endpoint and runs an ADOT Collector sidecar (digest-pinned, config injected via `AOT_CONFIG_CONTENT` — no custom image or SSM parameter). The sidecar is non-essential and memory-capped (64/256 MB) so it can never take down the serving container; scrape failures are visible in its `adot` log stream.
- **Endpoint privacy**: `GUARDIAN_METRICS_ADDR=127.0.0.1:9464` rather than the issue's `0.0.0.0` — Fargate `awsvpc` containers share one network namespace, so the sidecar scrapes loopback while nothing outside the task can reach the endpoint, independent of security-group contents. Never exposed via ALB or SGs.
- **Export**: awsemf (EMF log events into a Terraform-managed log group) under a per-stack namespace — `Guardian/Server` for the default stack, `Guardian-Prod/Server` for `guardian-prod` — so stacks in one account never mix. Dimension sets come from the closed label sets in `names.rs`; high-cardinality labels (route, method, operation) are rolled up, and only consumed rollups are exported. `retain_initial_value_of_delta_metric` is on so rare-event counters (first canonicalization failure) produce their first datapoint.
- **Dashboard** `<stack>-server`: request volume, error rate, latency, proposal/delta lifecycle, canonicalization health, storage/DB pools, Miden RPC, operator auth, ECS CPU/memory/tasks. SEARCH schema terms quote the namespace (hyphenated stack names are invalid unquoted).
- **Alarms** (9): HTTP 5xx rate and gRPC server-fault rate (shared error-code locals so dashboard and alarms can't disagree), latency, canonicalization failures (`error` **and** `partial` — passes with failed accounts report `partial`), metrics-missing (on the constant `guardian_build_info` heartbeat, traffic-independent), refresh-failures, refresh-stale (`DIFF` on the refresh timestamp — catches a hung refresher the failures counter can't), ECS CPU/memory (plan-time preconditions keep thresholds above the autoscaling targets). Wire notifications with `alarm_actions`.
- **IAM**: task role gets stream-level writes on the EMF log group only — no `CreateLogGroup`, no `PutMetricData`; Terraform stays authoritative for the group and retention.
[...]

PR #477, opened 2026-09-15
infra: manage the AWS Deploy workflow's GitHub OIDC roles in this repo

haseebrabbani:
This ports the GitHub Actions OIDC roles behind `.github/workflows/aws-deploy.yml` (#448) from oz-terraform into `infra/oidc.tf` and supersedes OpenZeppelin/oz-terraform#31.

### What's here
- `infra/oidc.tf`: bootstrap role in the root account (trusted by GitHub's OIDC provider, exact-match subject claims) and deploy role in the stacks account that it chains into, carrying a scoped inline policy limited to the ECR repositories, ECS services, task-definition families, and task roles of the stacks in `github_deploy_stack_names`. Opt-in via `github_oidc_enabled`; the roles are shared account-wide, so enable on exactly one stack per account. Defaults are OpenZeppelin's so adopting the existing roles is an in-place update.
- `aws.root_account` provider alias in `versions.tf`, reached via a named profile (`github_oidc_root_account_profile`) or an assumed role, with a precondition that it resolves to the account in `github_oidc_provider_arn`. The profile path is needed because the root-account `Terraform` role trusts only root-account principals.
- Outputs `github_oidc_role_arn` / `github_deploy_role_arn` — the values for the GitHub environment variables `ROLE_FOR_OIDC` / `ROLE_TO_ASSUME`.
- Docs: everything OIDC-specific lives in `docs/runbooks/github-oidc-deploy-roles.md` (how it works, OpenZeppelin's values, adopting the existing roles, adding environments/stacks, reuse from a fork); the generic deploy docs only link to it.

### Verified
Existing roles imported into the local `guardian-prod` state; `scripts/aws-deploy.sh plan` (Terraform 1.13) exits 0 and shows two OIDC changes: `aws_iam_role.github_oidc[0]` updated in place (`StringEquals` subjects, guardian-workflows claim dropped, default tags) and `aws_iam_role_policy.github_deploy[0]` created. The deploy role and the bootstrap role's assume policy are no-ops.

### Applying
The `guardian-prod` plan also contains unrelated drift from unapplied `main` (#442/#474/#475 observability and alarm resources, a new task-definition revision, RDS final-snapshot settings, a Cloudflare comment). Apply the OIDC change targeted unless that rollout is intended, then detach `AdministratorAccess` by hand — only after the apply, since the scoped policy must exist first:

    terraform -chdir=infra apply -state=terraform.guardian-prod.prod.tfstate \
      -target='aws_iam_role.github_oidc[0]' \
      -target='aws_iam_role_policy.github_deploy[0]' ...
    aws iam detach-role-policy --role-name GithubOIDCGuardianRole \
      --policy-arn arn:aws:iam::aws:policy/AdministratorAccess

Full sequence, including removing the resources from oz-terraform's `prod` workspace state afterwards, is in the runbook.

Follow-ups: close OpenZeppelin/oz-terraform#31 and OpenZeppelin/guardian-workflows#1 as superseded.


zeljkoX (approved):

PR looks good to me.

My main concern are docs.

We are explaining this optional infra, highly custom to OZ in generic docs. Let's move OIDC into separate document and just reference from generic docs:

The OIDC roles are only used by OpenZeppelin's own GitHub deploy, but the docs currently explain them in the generic deploy docs, where they read like a required step for everyone. I'd like to move all of it into one dedicated page and have the generic docs point to it.

New page: docs/runbooks/github-oidc-deploy-roles.md

Say at the top that this is the setup behind OpenZeppelin's AWS Deploy workflow, that it's off by default (github_oidc_enabled = false), and that it isn't needed for scripts/aws-deploy.sh. Then:

1. How it works: the bootstrap role → deploy role chain, what each role may do, and the requirement that each GitHub environment restricts deployment branches to main.
2. OpenZeppelin deployment:
   - a table of our values: guardian-prod as the managing stack, the two role names, the repo:OpenZeppelin/guardian:environment:* claims, and the guardian / guardian-prod deployable stacks;
   - the TF_VAR_* exports;
   - which output maps to ROLE_FOR_OIDC / ROLE_TO_ASSUME.
3. Adopting the existing roles:
   - the import runbook now in infra/README.md;
   - detach AdministratorAccess only after the apply has created the scoped policy, because detaching first leaves the role with no permissions;
   - finally, terraform state rm the four resources from oz-terraform.
4. Adding an environment or a stack: which variable to extend for each.
5. Reusing it for your own deployment:
   - the variables to override: github_oidc_subjects, github_deploy_stack_names, github_oidc_provider_arn (note that oidc.tf does not create the OIDC provider);
   - how to set it up with a separate root account vs. a single account (point github_oidc_root_account_profile at the same profile);
   - the OZ-specific parts of the workflow a fork must edit: the repository_owner guard, the environment choices, the GHCR image.

Generic docs: replace the explanations with a link

- infra/README.md:
  - drop the 8 github_oidc_* / github_deploy_* rows from the inputs table and the 2 rows from the outputs table;
  - cut the "GitHub Actions OIDC roles" section down to a few lines saying the roles are off by default and linking to the runbook.
- docs/SERVER_AWS_DEPLOY.md: replace the two OIDC bullets under "One-time setup per target" with one bullet linking to the runbook.
- infra/terraform.tfvars.example: shrink the block to # github_oidc_enabled = true plus a link.
- infra/oidc.tf header: say "exactly one stack per account" instead of naming guardian-prod, and point to the runbook instead of infra/README.md.
- docs/README.md: add the runbook to the Operations list.

The variable defaults can stay OpenZeppelin's, since they keep our import a no-op. The runbook's reuse section tells others what to override.

While updating, please also refresh the PR description:
- the null-validation commit it mentions is no longer in the PR;
- the -target list names aws_iam_role_policy_attachment.github_deploy_admin[0], which no longer exists;
- AdministratorAccess has to be detached by hand, not by the apply.

PR #478, opened 2026-09-16
feat(qualification): black-box end-to-end qualification suite

### LINKED ISSUE #432: Add black-box system E2E and multisig qualification
## Summary

Add automated end-to-end qualification for the assembled Guardian system: the release-style Docker image, Postgres, base clients, operator client, and both multisig SDKs.

Existing CI covers server modules, in-process E2E, Postgres implementations, and individual packages, but it does not prove that these components work together through real HTTP and gRPC boundaries. This is an umbrella issue; each phase may be delivered as a subissue.

## Phase 1: deterministic system E2E

Create a locally runnable harness and GitHub Actions workflow that:

- Builds the `server-runner` image with the `postgres` feature and current Git SHA.
- Starts Guardian and Postgres in isolated containers and waits for readiness with a bounded timeout.
- Verifies `/status`, including version, Git commit, and environment.
- Runs fixture account/proposal flows through:
  - `@openzeppelin/guardian-client` over HTTP.
  - `guardian-client` over gRPC.
- Verifies at least one structured API error.
- Restarts only Guardian and confirms account, state, and proposal data remain accessible.
- Captures test output, Guardian logs, and container state on failure.
- Always removes test containers and volumes.

Run this workflow on relevant pull requests, pushes to `main`, and `workflow_dispatch`.

## Phase 2: operator API E2E

Exercise `@openzeppelin/guardian-operator-client` against the real Postgres-backed server:

- Authenticate with a deterministic Falcon operator key.
- Establish and inspect a session.
- List accounts and fetch account details.
- Verify denial for an operator lacking the required permission.
- Verify allowlist hot reload without restarting Guardian.
- Log out and confirm session invalidation.
- Verify relevant audit events persist in Postgres.

## Phase 3: Miden multisig canaries

Run complete user-facing flows against the configured Miden test environment for both `miden-multisig-client` and `@openzeppelin/miden-multisig-client`:

- Create and register a multisig account.
- Create a proposal, collect threshold signatures, and execute it.
- Synchronize and verify state commitment and Guardian convergence.
- Cover Falcon and ECDSA signers.
- Cover online execution and offline export/sign/import.
- Cover `SwitchGuardian`.
- Verify Rust and TypeScript behavioral parity.

Because these tests depend on an external Miden environment, run them nightly and through `workflow_dispatch`, not as required pull-request checks. Expensive variants may rotate across runs, but disabled flows must be reported as skipped rather than passed.

## Acceptance criteria

- [ ] One command runs the deterministic system test locally.
- [ ] CI tests the real Docker image through public HTTP and gRPC ports.
- [ ] Both base clients complete their fixture flows.
- [ ] Postgres data survives a Guardian restart.
- [ ] Operator authentication, authorization, hot reload, logout, and audit persistence are automated.
- [ ] Nightly Rust and TypeScript multisig lifecycle coverage exists.
- [ ] Falcon, ECDSA, online, offline, and `SwitchGuardian` paths are covered.
- [ ] Readiness checks use timeouts rather than arbitrary sleeps.
- [ ] Failures preserve bounded diagnostics without exposing secrets or signed payloads.
- [ ] Third-party actions are pinned and workflow permissions are minimal.
- [ ] Local and CI execution is documented.

## Out of scope

- Automated Anvil/EVM proposal coverage.
- Package tarball/downstream-consumer qualification.
- Redesigning releases to gate publication on an exact qualified SHA.
- Load or production testing.

zeljkoX:
Closes #432.

Adds a black-box qualification suite that drives Guardian from the outside through both consumer SDKs, and is explicit about what a given run is entitled to claim.

## What it does

Two profiles:

| Profile | Chain | Trigger today | Intended |
|---|---|---|---|
| `deterministic` | none, Compose stack with an RPC stub | manual dispatch | required check on every non-doc PR |
| `live` | real Miden transactions from a treasury | nightly schedule, plus manual dispatch | unchanged, plus reviewer opt-in later |

Both SDKs run the **same scenario manifest**, so a Rust/TypeScript divergence surfaces as one leg passing and the other failing rather than as an untested gap. That is how most of the findings below were caught.

**43 scenarios** (28 live, 15 deterministic) over 48 actions, expanding to 67 SDK executions; 42 are required. 24 of the 28 live scenarios run on both SDKs. Coverage: account creation (public and private, Falcon and ECDSA), funding, note consumption, signer add and remove, removed-signer refusal, threshold change, per-procedure override and its survival across a threshold change, asset send, P2ID and timelocked P2IDE, offline export/import, cross-SDK handoff both directions, GUARDIAN rotation offline and online, producer-supplied proposals and the abandon control, account pausing against a live chain, registration scheme gates, error-envelope shape, operator session/denial/logout/allowlist reload, post-restart and post-upgrade durability, and negative controls.

## Layout

```
qualification/manifest/          scenario contract both drivers read
qualification/stack/             Compose stack, shell harness, redaction, teardown
crates/qualification-driver/     Rust driver, manifest validator, treasury
                                 accounting, run-result derivation, cross-SDK merge
packages/.../tests/qualification/   TypeScript driver
fixtures/qualification/          classification vectors both drivers are pinned to
```

## A run refuses to overclaim

This is the part worth reviewing hardest, because it is what makes a green run mean anything:

- an action with **no driver implementation fails closed** on a required scenario, so a coverage gap can never read as a pass
- an **empty required set** claims `none`, not `full`. `all()` is vacuously true on an empty iterator, and a live profile with no network resolves to exactly that set
- a **filtered** or **single-SDK** run cannot be upgraded to a full claim by the merge step, and produces no merged report
- the cross-SDK merge keeps the **worst verdict** per (scenario, SDK) across the restart and upgrade passes, with `Skipped` treated as the absence of a verdict, so neither pass can mask the other
- **environment-classified failures** do not block the conclusion, and the exit-code combiner still ranks product and setup failures above them
- the treasury **spend cap and lock** live in one directory shared by every spender, so they span the subprocess the TypeScript leg starts

## Telling the network apart from the product

A live run drives a public Miden network and a remote prover, neither under this repository's control. Three consecutive runs reported dropped connections and prover deadlines as `product` failures, which with the treasury armed and a nightly scheduled would page as a defect.

On the live profile only, a failure whose evidence points at the link is classified `environment`, which does not block the conclusion. The rule is the SDK clients' own transient-error classifier unchanged: permanent status evidence anywhere vetoes transient evidence anywhere. A quorum refusing an under-signed proposal still fails as `product`. Both drivers are pinned to `fixtures/qualification/environment-classification.json`, holding verbatim reasons from real runs on both sides of the line, and every run prints what it lost to the network.

## The completion rule is falsified, not just asserted

Completion is chain confirmation plus a canonical delta **at the proposal's own nonce**, and not "the proposal left the pending set", because canonicalization removes a discarded delta exactly as it removes a successful one. Matching on the commitment alone asks "is this account in a state some canonical delta explains", which an account whose delta was discarded satisfies just as well: it never moved, so it still agrees with chain and the previous delta still carries that commitment. Both drivers fail closed when the nonce cannot be read, because the read that supplies it happens before executing, and a failure there is exactly when the unbound comparison would wrongly confirm.

`live-custom-proposal-1of1-ecdsa` carries the negative control: it abandons a candidate that can never land, checks the discard is invisible to what a client reads by default, and then asks the completion helper itself about a delta that really was discarded. The candidate comes from the producer API, the one path that separates acknowledgement from submission: `prepare_custom_execution` pushes the delta to obtain the acknowledgement, and `submit_transaction` is a separate call. Every step is a supported public call. The control caught the unbound comparison in both drivers.

## Cross-SDK divergences this found

Each of these is one leg passing and the other failing on the same scenario, against the same GUARDIAN, minutes apart:

- **the TypeScript client does not surface a note its own account sent itself.** The Rust client lists the timelocked note as held and not yet consumable, which is the pair `live-p2ide-timelock-1of1-ecdsa` asserts. Neither a status listing nor an availability listing returned it on the TypeScript side three minutes after the transaction canonicalized, while its own output-note record was committed throughout. The TypeScript leg reads the landing from the sending side, which keeps the pair intact. Whether a TypeScript consumer can ever consume a note it sent itself is not answered here, and is worth answering
- **a proposal's metadata cannot be read back from GUARDIAN through the TypeScript client that created it.** `syncProposals` does fetch, but rebuilds the proposal from its own cached metadata when it has some, so a label GUARDIAN mangled would read back correctly on the proposing client and wrongly on every other one. The Rust client decodes GUARDIAN's answer either way
- **the proposer's signature is attached at creation in Rust and not in TypeScript**, so a 1-of-1 scenario that reaches execution without an explicit sign step passes on one SDK and fails on the other

## Product fixes included

- the Rust SDK could not change a threshold; `build_update_signers` now handles it
- `load()` returned an account whose reads came from a stale local store; reconciliation is now one shared rule so `load` and `syncState` cannot drift, and it refuses to overwrite state for an account that has transacted, and to adopt state that disagrees with chain
- Rust could not collect signatures off-channel; the offline-execution gate is gone and execution fetches the acknowledgement when the type requires one
- a grown operator-allowlist file read truncated through a Docker Desktop bind mount, and Guardian answered the transient read with a 500 and no retry
- `verify_endpoint_commitment` asked for a pubkey without a scheme, so an ECDSA account could never satisfy the check. **Landed separately as #479**, now in `main`

## Why the deterministic workflow is still dispatch-only

Its required scenarios pass. What is left is that it has not run green on a Linux runner, which cannot happen until this lands and the workflow exists on `main`, and the rule that a gate should be seen to go red on a real defect before anything depends on it. It has gone red several times, though each was found by adding or tightening a scenario rather than by catching a regression in existing coverage, so whether that clears the bar is a judgement for whoever owns the gate.

The live profile runs nightly against `qualification-devnet` and `qualification-testnet`, which exist, hold `QUAL_TREASURY_KEY` and are branch-locked to `main`. Qualifying a non-default ref still needs the treasury handoff, which is not built.

## State

Both profiles have a full-width green run behind them, not only per-scenario smoke runs:

| Run | Result |
|---|---|
| **Live, full matrix**, testnet | **52 legs, 52 passed**, conclusion `success`, claim `full`. 27 Rust, 25 TypeScript, ~30 minutes |
| **Deterministic, full** | 15 scenarios, 14 passed, 1 skipped (`operator-audit`, specified and unimplemented, so it is optional), claim `full` |
| **Upgrade from `v0.17.0`** | green end to end: seed phase on the old release, swap, target phase on the image under test |

The live run's leg durations are worth a glance, because they are the evidence it was not vacuous: the custom-proposal and abandon control takes 3m50s on Rust and 4m3s on TypeScript, which is the abandon quarantine and the completion deadline genuinely being waited out, while the six legs that finish in under a second are the three scenarios whose actions never touch the chain (`account-create account-register commitment-verify` and the cosigner recovery).

- 100 driver unit tests, 45 shell harness tests, 657 TypeScript tests, clippy clean, manifest valid, TypeScript driver typechecked in CI
- every live scenario was verified on testnet, and the ones whose assertions could plausibly be vacuous were also verified to **fail** when the property under test is removed: the online rotation, the P2IDE timelock, the merge fold, and the empty-store adoption check
- the offline migration passes on both SDKs against the stack's real second GUARDIAN
- the upgrade pass runs automatically when a change touches `crates/server/migrations/`, and is refused on the live profile, where it would fund every scenario twice

## Summary by CodeRabbit

* **New Features**
  * Added deterministic and live qualification workflows covering Guardian, multisig, operator, migration, SDK interoperability, and network scenarios.
  * Added qualification reporting, manifest validation, treasury setup, funding controls, diagnostics, and secret scanning.
  * Added account-wide threshold updates and broader offline proposal signing and execution support.
  * Added safer Guardian account-state reconciliation with on-chain commitment checks.
* **Bug Fixes**
  * Improved allowlist loading resilience during temporary file replacement.
  * Added regression coverage ensuring incomplete multisig signatures remain rejected.
* **Documentation**
  * Added qualification, treasury, multisig, compatibility, and troubleshooting guidance.


haseebrabbani (approved):
approved, some comments:

1. `--upgrade-from` + `--scenario` (the pairing the usage text recommends) passes `--filtered` twice to the seed phase: once from `DRIVER_ARGS` (run.sh:414) and once from `SEED_SCENARIOS` (run.sh:436). clap rejects it (`the argument '--filtered' cannot be used multiple times`, confirmed with the binary), the seed is skipped with a "note", and `det-restart-durability` then fails as a *product* failure on an empty database. CI never passes `--scenario`, so only local users hit it.
2. The live classifier applies the SDK's fallback vocabulary (`unavailable`, `timeout`, …) to the driver's *composed* reason (environment.rs:66). live.rs:787 writes `delta history unavailable: {error}`, so a GUARDIAN 500 on delta history classifies as `environment` and stops blocking the nightly conclusion. Probed: `…delta history unavailable: gRPC Internal server error` and `…timeout waiting for quorum…` both come back environmental. Classify on the caught error chain, or keep fallback words out of driver-composed text.

PR #479, opened 2026-09-16
fix(multisig-client): bind the endpoint commitment check to the account's scheme

zeljkoX:
Split out of #478 so it can be reviewed and land on its own. The change is identical in both branches; whichever merges first, the other resolves cleanly.

## The bug

`verify_endpoint_commitment` called `client.get_pubkey(None)`.

Guardian holds **one acknowledgement identity per signature scheme**. Asked without a scheme, it returns the default. So the check compared an account's stored guardian commitment against whichever identity happened to be the default on that server, not against the identity the account was actually registered with.

For a Falcon account on a default-Falcon Guardian, that is accidentally correct, which is why it went unnoticed.

For an ECDSA account it can never succeed. The account's guardian slot holds the ECDSA acknowledgement commitment, the call returns the Falcon one, and they never match. The observable effect is that migrating an ECDSA account to a new Guardian endpoint is refused whatever endpoint is passed, with an error that points at the endpoint rather than at the scheme.

## The fix

`scheme: SignatureScheme` is a **required** parameter, not an `Option`. Every call site must state which identity it means, and the compiler rejects a new caller that does not. All three callers already had the account's key manager in scope:

| Call site | Source |
|---|---|
| `client/offline.rs` | `self.key_manager.scheme()` |
| `client/helpers.rs` | `self.key_manager.scheme()` |
| `transaction/builder.rs` | `key_manager.scheme()` |

## Scope

`verify_endpoint_commitment` is `pub(crate)` in a private module, so this is not a public API change and no README or crate-docs update is owed.

## Testing

No unit test is added here, deliberately. The interesting property is that a caller cannot omit the scheme, and that is now enforced by the type signature, which is stronger than a test asserting it. I wrote a test that inspected the source for the call shape, realised its own assertion string satisfied the match, and deleted it.

The behaviour is covered end to end instead: the qualification suite in #432 runs guardian migration on ECDSA accounts against a real Guardian on both SDKs, which is where this was found and which fails without the fix.

Refs #432

## Summary by CodeRabbit

- **Bug Fixes**
  - Guardian endpoint validation now uses the configured signature scheme consistently.
  - Guardian endpoint switches and offline transaction proposals correctly verify commitments against the appropriate guardian public key.
  - Improves validation reliability for configurations using non-default signature schemes.


haseebrabbani (approved):
lgtm

